Issue #550, opened 2026-07-20
Release-day checklist — 2026-07-28 (target: 2.0.0-rc.1 on the 2.x ladder)


mapyr:
Release-day runbook for **2026-07-28** (MCP `2026-07-28` spec RC day). **Target corrected:** the day's release is **`2.0.0-rc.1`** on the `2.0.0a1 → b → rc → GA` ladder — **not** v1.6.0 (already shipped and live). GA (`2.0.0`) follows once rc soaks.

> Reconstructed runbook — prior verbal agreement wasn't in the working context; adjust freely.

## Ladder position (context)
- `2.0.0a1` — staked now from `mcp2`, pre-SDK-v2-migration (config: #549 / this milestone's PR). TestPyPI.
- `2.0.0b1…` — after the **SDK v2 migration** (#547 is the gating spike; migration is its own workstream). TestPyPI.
- `2.0.0-rc.1` — **28.07 target**. TestPyPI, feature-complete against the `2026-07-28` spec.
- `2.0.0` GA — via `release-please` on `main` per ADR-009 Decision 1, to prod PyPI, after rc soak.

## Gate: must be true before cutting rc.1
- [ ] **SDK v2 migration landed on mcp2** (#547 catalog worked through: FastMCP→MCPServer, `mcp_types`, `request_ctx`→`ServerRequestContext`, httpx, Tasks extension). Full suite + live tiers green.
- [ ] **Spec alignment (#310)** first slice done: WS-1 stateless transport verified against `2026-07-28`; deprecated-surface handling (roots/sampling/logging transition window) confirmed.
- [ ] **Cross-generation compatibility harness green** (benchmarks `compat/`): legacy (`initialize`+`Mcp-Session-Id`) AND modern (`server/discover`+`Mcp-Method`/`Mcp-Name`) both served by one gateway; version negotiation; Tasks relay-only asserted as a clean `TaskRelayNotSupported` (not a hang).
- [ ] **Compatibility matrix (#453)** published with owner + ranges (core × operator × charts × K8s × MCP protocol).
- [ ] Fingerprinting/shadow-MCP signatures (if implemented) target `server/discover` + `Mcp-Method`/`Mcp-Name`, **not** the deprecated `initialize` handshake (operator#53 note).

## ADR state
[...]

PR #609, opened 2026-07-27
fix(core): cap the mcp SDK pin at the v1 line

### LINKED ISSUE #561: mcp SDK pin has no upper bound — `--pre` installs of the v1-SDK line pull mcp v2 beta and crash on import
Found by testing the published artifact (compat gate D) in a clean venv.

## Symptom
```
pip install --pre mcp-hangar==2.0.0a1   # or: uv pip install --prerelease=allow mcp-hangar==2.0.0a1
```
resolves **`mcp==2.0.0b2`** (SDK v2 beta) as the transitive dep, and the gateway crashes on import:
```
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```
`mcp-hangar --version` fails; the process never starts.

## Root cause
`pyproject.toml` pins `mcp>=1.28.1` with **no upper bound**. The 2.0.0a1 code targets the **SDK v1** surface (`mcp.server.fastmcp`, etc.), but under prerelease resolution (`--pre` / `--prerelease=allow`) `mcp` floats up to `2.0.0b2`, which removed that surface (#547 catalog). Since 2.0.0a1 is itself a prerelease, users install it *with* `--pre` — so this is the default install path for the alpha, and it's broken.

## Verified
- `--pre` → mcp 2.0.0b2 → crash.
- Forcing `mcp<2` → mcp 1.28.1 → `mcp-hangar 2.0.0a1` starts; compat harness (benchmarks#4) passes.

## Fix
- Cap the pin for the **v1-SDK line**: `mcp>=1.28.1,<2` on `mcp2` (until the #547 SDK v2 migration lands; the `b`-series then re-widens/moves the pin to `mcp==2.0.0bN`).
- 2.0.0a1 itself is immutable on PyPI — mitigated in its release notes (install with `"mcp>=1.28.1,<2"`).
- Add a **published-artifact smoke** (gate D, #550) to CI so a broken prerelease install is caught before release, not after.

Refs: #547, #550 (gate D), benchmarks#4, v2.0.0-alpha.1.

mapyr:
## Why

**Time-critical.** This branch serves the v1 SDK surface (`mcp.server.fastmcp`), which SDK v2 removed, but the dependency is declared `mcp>=1.28.1` — **no upper bound**. Resolution follows the SDK across the major and the gateway dies at import:

```
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

- Under `--pre` / `--prerelease=allow` this happens **today**: a clean venv resolves `mcp 2.0.0b2`.
- It starts happening on a **plain `pip install`, no flag**, the moment `mcp 2.0.0` ships. Upstream targets **2026-07-28**, so this is days away.

1.6.1 is already published with the uncapped pin, and a published wheel cannot be edited — this is one of the few defects a release makes permanent for everyone who installed it.

Closes #561

## Reproduced, then fixed

Clean venv, prerelease resolution on:

| constraint | resolves `mcp` to |
| --- | --- |
| `mcp>=1.28.1` (before) | **2.0.0b2** → import crash |
| `mcp>=1.28.1,<2` (after) | **1.28.1** |

Installing this branch's own metadata: `mcp-hangar 1.6.1` + `mcp 1.28.1`.

Incidentally, my local checkout demonstrated the consequence while I worked: with `mcp 2.0.0b2` in the venv, **55 of this branch's test modules fail collection** on exactly that `ModuleNotFoundError`. That is what a user gets.

## What

- `mcp>=1.28.1,<2`, with a comment saying why the cap is deliberate and that `mcp2` (the v2 line) pins `mcp==2.0.0bN` exactly and must not follow this one.
- `uv.lock` regenerated so the recorded specifier matches (`{ name = "mcp", specifier = ">=1.28.1,<2" }`); resolved `mcp` stays 1.28.1.
- A metadata test asserting the bound **exists** and **actually excludes 2.x** — a bound that is merely present is not the property that matters. It rejects `2.0.0`, `2.0.0b2` and `2.1.0`, and still accepts `1.28.1`. Both assertions fail without the change.

## How tested

`pytest tests/unit/test_sdk_pin_bounds.py` → 5 passed; 2 fail without the cap.

I could **not** run the full suite locally: my checkout's venv holds `mcp 2.0.0b2` for the `mcp2` work, so this branch's modules do not import there — the bug itself. CI runs the suite against the regenerated lock, which is the environment that matters for this change.

## Risk and rollback

Low; revert the single commit. The cap only forbids a major this code cannot run on. Anyone who genuinely wants Hangar on SDK v2 wants the `mcp2` line, which is pinned there.

## Follow-up

#550's gate D (published-artifact smoke in a clean venv) is the check that would have caught this before release rather than after. Still unbuilt; worth doing before the next tag on either line.

## Agent metadata

- [x] LOC declared upfront: ~90 (10 pyproject, 4 lock, 72 test, 6 changelog)


mapyr:
Closed immediately after opening: this PR was created against a **pre-existing remote branch of the same name**, not the work intended for it.

`fix/cap-mcp-sdk-pin` already existed on the remote from an earlier, abandoned attempt at #561 (`e770066`), branched before the 1.6.1 release. As opened, this PR would have merged that stale lineage into `main` — it lacks `5a24f20` (release 1.6.1), `969b996` (MCPEgressPolicy mode) and others, so merging it risked reverting shipped work.

Superseded by a PR from a distinct branch, based on current `main`.

PR #698, opened 2026-08-03
ci(ci): run the decision-path coverage gate

mapyr:
## Why

Wires the decision-path coverage gate from #697 into CI. Without this the gate
exists and passes, but nothing runs it.

## What changed

One job, `decision-coverage`, in `ci-core.yml`.

A **separate job** rather than a step inside `test`:

- `test` is a 3-version matrix — running the gate three times triples the flake
  surface for one signal
- `test` passes no `-m` filter, so its selection is not the one the floors were
  measured against; this job pins its own so the number is reproducible
- a named job can be made a required check without making the whole matrix
  required

Branch mode is **not** passed as a flag here. It comes from
`[tool.coverage.run]`, and the gate refuses a statement-only report outright — so
the workflow and the config cannot drift into measuring different things.

## How tested

The payload was verified in #697, by exit code rather than by eyeballing output:
a module below its floor exits 1, a report produced without branch mode exits 2,
a floored module missing from the report exits 2, and a clean report exits 0. The
same command this job runs was executed locally against the merged tree — 5334
tests, all 29 floors held.

What could not be verified before merge is the job definition itself running on a
GitHub runner. It is 33 lines invoking a script whose behaviour is pinned above;
the first run on `main` is the real test.

## Ordering

Merge **after** #697 (now merged). On its own this references a script and a
config section that would not exist.

## A correction, since it is written into this branch's commit message

The commit says checks failed to dispatch on the #697 branch *because* it carried
a workflow-file change. **That explanation is wrong** — this PR changes only
`.github/workflows/ci-core.yml` and its checks dispatched normally.

What actually happened on #697 is unexplained: for roughly 45 minutes no
workflows fired on that branch (not `ci-core`, not `pr-title`, which has no paths
filter), through a close/reopen and a `synchronize` push, while `main` was
running workflows normally in the same window. Merging `main` into the branch
produced an event that was delivered. Given the GitHub 503s and API timeouts seen
repeatedly today, transient event-delivery loss is the likelier cause than any
repo policy.

Keeping workflow changes in small separate PRs is still reasonable, but it is a
preference here, not a workaround for a known rule.

### DIFF
--- a/.github/workflows/ci-core.yml
+++ b/.github/workflows/ci-core.yml
@@ -83,6 +83,39 @@ jobs:
           flags: core
           fail_ci_if_error: false
 
+  decision-coverage:
+    # A SEPARATE job, not a step folded into `test`, on purpose:
+    #  * `test` is a 3-version matrix; running the gate three times triples the
+    #    flake surface for one signal.
+    #  * `test` passes no `-m` filter, so its selection is not the selection the
+    #    floors in [tool.decision_coverage.floors] were measured against. This
+    #    job pins its own so the number is reproducible.
+    #  * a dedicated job name can be made a required check without making the
+    #    whole matrix required.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Install dependencies
+        run: pip install -e ".[dev]"
+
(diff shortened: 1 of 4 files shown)


mapyr:
Superseded by #701. This branch was cut from the #697 working branch rather than `main`, so after #697 landed as a squash it re-added the script and floors already on `main` and conflicted. #701 carries the workflow change alone.

Issue #753, opened 2026-08-05
epic: wire the event-sourcing spine the platform already advertises (ADR-002 remodel)


mapyr:
The audit below is verified against the tree at `a81fac7`, not inferred.

## Why

`ADR-002` (**Accepted**, 2026-04-17) declares event sourcing "the primary persistence mechanism for domain aggregates". The website says security-relevant events are "emitted onto an append-only, event-sourced chain (the same event-sourcing spine the rest of the platform is built on)", and `README.md:70` advertises "an event-sourced trail exported to SIEM".

None of the durability half of that is true today:

- `publish_to_stream()` / `publish_aggregate_events()` — the **only** writers — have **zero** production call sites. Verified three ways: grep, an AST pass over every method call on an `event_bus` object (39 `publish()`, 0 `publish_to_stream()`), and a check for dynamic dispatch. They appear in two test files and nowhere else.
- `data/events.db` was created 2026-07-16. **All four tables hold 0 rows.**
- Bootstrap refuses to start if that path is not writable and reports `durable=True` to the health endpoint — for a store nothing writes to.
- `ADR-002` names `domain.model.event_sourced_mcp_server.py`, deleted in #749 as never wired; `FileEventStore`, which has zero uses outside its own module; and `EventStoreSnapshot`, which is only re-exported.
- Actual aggregate persistence is `SQLiteMcpServerConfigRepository` — state snapshots, the mechanism ADR-002 says it rejected.

What **is** true and stays: OTLP audit export gated on `OTEL_EXPORTER_OTLP_ENDPOINT`, and the CEF / LEEF / syslog / JSON-lines exporters gated on `MCP_COMPLIANCE_FORMAT`. Both are wired exactly as documented.

## The good news

This is a pipe that is laid but not connected, not a rewrite:

- `domain/model/aggregate.py` already provides `AggregateRoot._record_event` / `collect_events`; `McpServer` records events in **18** places.
- **12 drain points** already do `for event in agg.collect_events(): bus.publish(event)`. They need to append to a stream instead.
- The serializer (registry derived from the class hierarchy), the upcaster chain, optimistic concurrency and the SQLite store all exist and are tested.
- The hexagon is enforced by import-linter; the debt ledger is down to **9** edges.
[...]

PR #761, opened 2026-08-05
refactor(core): remove eight event classes nothing emitted

mapyr:
**Stacked on #760** — that PR adds the baseline this one shrinks. Merge #760 first; GitHub will retarget this to `main`.

## Why

Eight event classes had **no producer and no consumer anywhere in the tree**: vocabulary for features nobody built. The gate in #760 is what made that list countable rather than anecdotal, and this is the half of the debt that can actually be paid — the other ten stay for reasons given below.

`CatalogItemApproved`, `CatalogItemDeprecated`, `CatalogItemPublished`, `CatalogItemRejected`, `ToolSchemaChanged`, `ToolSchemaDriftDetected`, `BehavioralModeChanged`, `CapabilityDeclarationMissing`.

## What

Deletes those eight, their exports and their tests; shrinks the `[tool.event_contracts] unused_events` baseline from 18 to 10; adds the `UPGRADE.md` section.

**Breaking, and exactly how far.** Only code that *imports one of those names*. Nothing else can be affected, and that is a statement with a proof rather than a hope:

- no deployment can have *received* one, because nothing has ever constructed one;
- a handler registered against any of them has therefore never been called, so deleting it changes no behaviour;
- no stream can contain one, so no event store needs migrating.

Minor + `UPGRADE.md`, per `AGENTS.md`.

**What stays, and why the distinction matters.** Ten unemitted events remain on the baseline:

- **the five discovery events** — the feature is live and wired into `cqrs.py`; its orchestrator simply publishes nothing. That is a *missing emitter*, not dead vocabulary, and deleting them would decide by cleanup that discovery will never report anything. Tracked in #762.
- **four quarantine events and `PolicyPushRejected`** — these touch the operator.

Their tests go with them: they constructed the class and asserted on the result, which — once nothing else uses the class — tests only that a dataclass is a dataclass.

## How tested

`5682` unit tests pass. Dead-symbol baseline **shrinks** 37 → 32 exported-unreferenced and is locked at the smaller number. `ruff` (pinned 0.16.1), `lint-imports` **KEPT**, `markdownlint` clean on the upgrade note.

**One threshold moved, and it is worth reading.** `test_the_registry_is_not_trivially_small` asserted `len(EVENT_TYPE_MAP) >= 100`; the tree now holds 93. That check is a **floor against the registry coming up empty**, not a census: coverage is guaranteed by `test_every_event_class_can_be_deserialised`, which derives the expected set from the class hierarchy and needs no maintenance when events come or go. Lowered to 50, with the reasoning written where the number is so the next reader does not meet a magic constant and guess.

## Risk and rollback

The removed classes cannot appear in any stream or reach any handler, so the runtime blast radius is empty; the exposure is an `ImportError` for a downstream that imported a name it could never have used. Revert is a single commit.

Refs #753


mapyr:
Re-targeted as a fresh PR against `main`.

This PR is marked MERGED and its content **never reached `main`**: it merged into `test/event-producer-consumer-gate` at 11:33:05, 26 minutes after that branch was squash-merged into `main` at 11:07:33. A squash collapses the branch into a new commit on `main` and leaves the branch itself alive, so anything merged into it afterwards goes nowhere.

The eight classes are still in `main`; verified by grep, not inferred. The original commit `d3aeac8` is cherry-picked onto current `main` in the replacement PR.

Issue #790, opened 2026-08-06
epic: high availability -- the work, in dependency order


mapyr:
The work to deliver the design in #789. Evidence in #778, storage groundwork in
#779 (done, #780–#787), the flow analysis in #788.

Ordered by dependency, not by importance. **Phase 0 is repair that stands on its
own** — every item there is a defect today, at one replica, and none of it is
wasted if the shape in #789 is later revised.

Each item names how it is verified, because "wired but silent" is the failure
this codebase produces most reliably and every phase here has room for it.

---

### 0.1 Events carry the instance that produced them
`DomainEvent` has exactly two fields, `event_id` and `occurred_at`. There is no
producer identity, so three replicas write rows nobody can tell apart, and a
tailer cannot skip its own appends — which is the correctness gap in my review of
#789, not merely an audit inconvenience.

An instance id fixed at startup, on the base event, carried through
serialization.

- **Watch out:** events are persisted, so this is schema evolution. The upcaster
  chain exists and this must ride it — an old row without the field must still
  deserialize. ADR-018 already says an event's shape is a compatibility surface.
- **Verified by:** a stored pre-change row replays; two buses with different
  instance ids produce distinguishable rows; the field survives a round trip.
- **Size:** small, but touches the serializer. **Blocks 0.4, 2.1.**

### 0.2 Commit-ordered log
`global_position` is a `BIGSERIAL`; allocation order is not commit order, and the
PostgreSQL adapter's own docstring admits it and advises tolerating reordering —
advice a monotonic cursor cannot take. `dispatch_pending` survives only because
it runs once at startup, after the tail has settled.

**Do the cost work before choosing the mechanism.** #789 proposes a counter row
in the append transaction and calls the cost theoretical on the grounds that the
log carries management events. It does not: `ToolInvocationCompleted` is recorded
on every tool call and persists since #774, so that counter would serialize every
invocation in the cluster behind one row lock. Either measure it against a real
invocation rate, or make `xmin` fencing (`pg_current_snapshot()`) the design
rather than the escape hatch.

- SQLite needs no change — a single writer makes allocation order commit order —
  but it needs the note, so the two backends are known to agree.
- **Verified by:** concurrent appenders, a tailer with a monotonic cursor, and an
  assertion that nothing is skipped. This is the one item where a test that
  merely passes proves nothing; it has to reproduce the race first.
- **Size:** medium, with a benchmark attached. **Blocks 2.1.**

[...]

PR #808, opened 2026-08-06
feat(core): a hangar cluster requires PostgreSQL, and says so at startup

mapyr:
Found by running three replicas on SQLite in a real cluster, after #790's cross-cluster testing.

## Why

Several replicas on a file-backed backend **do not collide**. Each pod gets its own file. Each grants itself its own lease — the SQLite adapter always grants, and that is correct, because a file admits one writer. Each runs the management loops. Each holds its own fleet.

They never disagree, because they cannot see each other. So every health check stays green while the deployment has as many fleets as it has pods.

Measured before this existed, three replicas, `persistence.backend: sqlite`:

```
hangar-sqlite-...-84mhz: manages_fleet= True   coordinates_with_peers= True
hangar-sqlite-...-nrgdn: manages_fleet= True   coordinates_with_peers= True
hangar-sqlite-...-w7dpt: manages_fleet= True   coordinates_with_peers= True
```

Three leaders, and the API claiming coordination on all three — because I had written `coordinates_with_peers = keeper is not None` in #803. **A keeper is not coordination; a shared backend is.**

## What

**A `coordination:` block is the statement that these replicas are meant to be one gateway**, and it is refused on storage they cannot share:

```
this gateway is configured as part of a cluster (`coordination:`), and the 'SqliteBackend'
storage backend is local to one process. Replicas that cannot share storage are not a
cluster: each would hold its own fleet and its own lease, and they would never notice each
other. Use `persistence.backend: postgresql`, or remove the `coordination:` block to run
this as a single gateway.
```

The question is asked on **the axis the operator controls**, not by sniffing the environment. An earlier draft refused whenever `KUBERNETES_SERVICE_HOST` was set; that is the wrong axis — a thousand pods each with their own storage are a thousand gateways, which is a legitimate thing to run and nobody's business but the operator's. What is not legitimate is calling them one gateway.

Two consequences fall out of it:

- **A backend that cannot be shared gets no lease keeper at all.** There is no peer to coordinate with, so a keeper would grant itself the lease every time and report that it coordinates — which is precisely what produced the three-leader reading above.
- **`GET /api/system` stops claiming what it does not have.** `storage_is_shareable` is reported, and `coordinates_with_peers` is true only when it is. `manages_fleet: true` on a file-backed gateway stays as it was, because it is not a lie: it really is the manager of the only fleet it can see. Paired with the other two fields, an operator can read what that means.

A backend that has not declared itself shareable counts as unshared. One whose adapters have not been examined for the question is not one to assume the answer for.

## How tested

Both directions on the cluster where the problem was found. With `coordination:` and SQLite, three replicas crash-loop with the message above. Without it, the same three start and report honestly: `manages_fleet: True, coordinates_with_peers: False, storage_is_shareable: False`.

`tests/unit/test_replicas_mean_postgresql.py` (11): each backend's answer to whether it can be shared, including that an unknown one counts as unshared; a cluster on unshared storage is refused, with both ways out in the message; a shared backend never is; not asking for a cluster is always fine; a file-backed backend gets no keeper while a shareable one does; and the three API fields, including that managing its own island is still reported truthfully.

Two existing tests needed a deliberate edit: they built a keeper from a SQLite backend, which now correctly returns None, and what they are actually about is the keeper rather than which backend deserves one.

Full run green: 6485 passed, unit + integration. Gates hold: ruff 0.16.1, mypy (5 pre-existing), import-linter, dead symbols 42/32.

Note for merge order: this touches `init_lease_keeper`, which #805 also edits. Whichever lands second will want a look at that function — the two changes are in adjacent branches of the same early-return.

https://claude.ai/code/session_014aPr4VowiQT2v4QyJiFzzW


mapyr:
Rebased onto main after #804–#807 merged. Two changes beyond a mechanical replay, both from the same collision with #805:

**`init_lease_keeper` had to be resolved by hand.** #805 added `set_local_mode_policy(...)` to the same function this PR adds the shareability gate to. A naive resolution takes one side and silently reverts the other — taking this branch's version drops the local-mode wiring #805 just landed. The resolved function keeps both, and the new early return sets `set_local_mode_policy(None)` for the same reason the no-backend branch does: a policy left in the module by a previous bootstrap in the same process must not survive into a gateway that has no follower to be.

**One of #805's tests asserted a rule this PR changes.** `test_selecting_a_backend_makes_local_modes_leader_owned` created a SQLite backend and asserted local modes became leader-owned. That was true when *selecting a backend* was the question; here the question is whether the backend is **shareable**, and a file-backed one is standalone, so its `subprocess` servers stay available. It is now two tests — a shareable backend gets the policy, a file-backed one keeps every mode — which pins both halves of the rule instead of one.

Verified on the rebased commit locally before pushing: 6539 passed, ruff + format clean, import contract kept, dead-symbol baseline 42/32. CI is green on `ad7995c`, including decision-coverage.

Issue #865, opened 2026-08-10
Test that REST and MCP agree on the same Host header


mapyr:
Sub-task of #859.

## What

A test that pins REST and MCP to the *same* answer for the same `Host`, because the bug was precisely that they disagreed.

## Acceptance

With `MCP_TRUSTED_HOSTS` set to a list that includes a Service-style DNS name, assert the full matrix rather than one case:

| `Host:` | `POST /mcp` | `GET /api/system/` |
|---|---|---|
| an allow-listed name | 200 | 200 |
| an allow-listed name with a port | 200 | 200 |
| a name **not** on the list | 421 | 400 |

The failing row today is "an allow-listed name" against `/mcp`, which returns `421`. A single-endpoint test would not have caught it — the REST side was always correct.

Issue #864, opened 2026-08-10
MCP_TRUSTED_HOSTS entries need expanding to the SDK's host:port matching


mapyr:
Sub-task of #859.

## What

The two allowlists do not match the same way, so handing `MCP_TRUSTED_HOSTS` straight to the SDK is not enough.

Starlette's `TrustedHostMiddleware` and `trusted_hosts.host_is_trusted()` compare the host **with the port stripped**. The SDK's `TransportSecurityMiddleware._validate_host` compares the raw `Host` header against its entries, with only a `base:*` suffix form:

```python
if host in self.settings.allowed_hosts:
    return True
for allowed in self.settings.allowed_hosts:
    if allowed.endswith(":*"):
        base_host = allowed[:-2]
        if host.startswith(base_host + ":"):
            return True
```

Measured on a running 2.5.1 replica, the effective allowlist is `127.0.0.1:8080` and `localhost:8080` — **a bare `localhost` is refused**, which is why an nginx `upstream-vhost: "localhost"` does not work and `"localhost:8080"` does.

## Acceptance

Operators write hostnames, not host:port pairs. An entry `example.internal` in `MCP_TRUSTED_HOSTS` should admit `example.internal` **and** `example.internal:<any port>` on the MCP endpoint, i.e. each entry expands to `entry` plus `entry:*`.

Wildcard entries (`*.example.com`) and `*` need the same treatment or an explicit statement that they are unsupported on this surface.

Issue #863, opened 2026-08-10
Feed the trusted-hosts allowlist into the MCP endpoint's transport security


mapyr:
Sub-task of #859.

## What

`server/lifecycle.py` mounts the MCP app with the SDK's defaults:

```python
mcp_app = mcp_server.streamable_http_app()
```

so `transport_security` is `None` and `host` defaults to `127.0.0.1`. The SDK builds its DNS-rebinding allowlist from that and rejects everything else with `421`.

Pass the configured allowlist instead, from `trusted_hosts.trusted_hosts()` — the same source `api/router.py` already feeds to Starlette's `TrustedHostMiddleware`.

## Acceptance

- `POST /mcp` with a `Host` in `MCP_TRUSTED_HOSTS` succeeds.
- `POST /mcp` with a `Host` outside it still fails — this is not "turn the guard off".
- `MCP_TRUSTED_HOSTS=*` opts out entirely, matching `TrustedHostMiddleware`'s own escape hatch.
- `fastmcp_server/factory.py:163` builds the same app; whichever wiring is chosen has to cover both call sites or they drift.

## Note

`trusted_hosts.py` already claims this is how it works:

> One allowlist, read by every layer that needs it: `server.api.router` configures Starlette's `TrustedHostMiddleware` from it, `fastmcp_server.asgi` checks the MCP endpoint's Host against it …

This makes the docstring true.

PR #871, opened 2026-08-10
fix(core): read the trusted-hosts allowlist on the MCP endpoint too

### LINKED ISSUE #859: MCP_TRUSTED_HOSTS does not reach the MCP endpoint, so /mcp is unreachable through a Service or Ingress
## Summary

`MCP_TRUSTED_HOSTS` governs the REST API but not the MCP endpoint. `lifecycle.py` mounts the Streamable HTTP app with no `host=` and no `transport_security`, so the SDK builds its DNS-rebinding allowlist from the default `127.0.0.1` and answers `421 Invalid Host header` to every other name — including the Service DNS name and any Ingress host.

The practical effect: **the MCP surface cannot be reached through a Kubernetes Service or an Ingress**, which is the shape of every multi-replica deployment.

Reproduced against `ghcr.io/mcp-hangar/mcp-hangar:2.5.1` with `mcp==2.0.0`.

## Cause

`server/lifecycle.py`:

```python
mcp_app = mcp_server.streamable_http_app()
```

`streamable_http_app(..., transport_security=None, host="127.0.0.1")` — the SDK then enforces host:port entries derived from that bind host. `MCP_TRUSTED_HOSTS` reaches only `api/router.py`:

```python
app.add_middleware(TrustedHostMiddleware, allowed_hosts=trusted_hosts())
```

`trusted_hosts.py` states the opposite in its own module docstring:

> One allowlist, read by every layer that needs it: `server.api.router` configures Starlette's `TrustedHostMiddleware` from it, **`fastmcp_server.asgi` checks the MCP endpoint's Host against it** … A security allowlist that three call sites disagree about is worse than either answer consistently applied: the disagreement is what an attacker gets to stand in.

Two call sites do disagree, and the MCP one is not reading it.

## Reproduction

One replica, with every name explicitly listed:

```
MCP_TRUSTED_HOSTS=localhost,127.0.0.1,::1,mcp-hangar,mcp-hangar.hangar,mcp-hangar.hangar.svc,mcp-hangar.hangar.svc.cluster.local,hangar.lab.local
```

Same request body, only the `Host` header varying, issued from inside the pod against `127.0.0.1:8080`:

| `Host:` | `POST /mcp` | `GET /api/system/` |
|---|---|---|
| `127.0.0.1:8080` | 200 | 200 |
| `localhost:8080` | 200 | 200 |
| `127.0.0.1` | **421** | 200 |
| `localhost` | **421** | 200 |
| `mcp-hangar.hangar.svc.cluster.local:8080` | **421** | 200 |
| `hangar.lab.local` (via Ingress) | **421** | 200 |

Note that the allowlist entries carry the port, so even a bare `localhost` is refused.

**Expected:** the MCP endpoint honours `MCP_TRUSTED_HOSTS`, as the module docstring says it does.

## Workaround

Make the proxy lie about the Host:

```yaml
nginx.ingress.kubernetes.io/upstream-vhost: "localhost:8080"
```

This works, and it is not a good place to leave the security boundary: the allowlist that recipe 23 tells operators to replace has no effect on the surface clients actually use.

## Adjacent, worth a line in the docs either way

A Streamable HTTP session lives in one replica's memory, so a round-robin Service or Ingress sends the second request of a session to a pod that never saw the first. It needs `sessionAffinity: ClientIP` on the Service and something like `upstream-hash-by: $remote_addr` on the Ingress. Cookbook 25 says serving works on every replica — true — and does not mention that a session has to keep reaching the same one.

mapyr:
## Why

`MCP_TRUSTED_HOSTS` governed the REST API and not the MCP endpoint, so one process gave two answers to the same `Host` header — and the surface clients actually use was the one that said no. Behind a Kubernetes Service or an Ingress, `/mcp` was unreachable.

Closes #859, #863, #864, #865

## How tested

- `pytest tests/unit tests/integration` — 6729 passed, 29 skipped.
- Behavioural check against a running gateway with `MCP_TRUSTED_HOSTS=localhost,127.0.0.1,hangar.example.test`:

| `Host:` | `/mcp` before | after |
|---|---|---|
| `127.0.0.1:8097` | 200 | 200 |
| `hangar.example.test` | 421 | **200** |
| `hangar.example.test:8097` | 421 | **200** |
| `evil.example.test` | 421 | 421 |

The last row is the point: this feeds the guard a different list, it does not turn it off.

Originally measured on a released 2.5.1 replica in a kind cluster, where `mcp-hangar.hangar.svc.cluster.local:8080` answered 421 on `/mcp` and 200 on `/api/system/` with the name listed explicitly.

## What

- `mcp_transport_security()` builds the SDK's `TransportSecuritySettings` from `trusted_hosts()`, and both serving paths (`ServerLifecycle.run_http`, `MCPServerFactory`) pass it to `streamable_http_app()`.
- Each entry is expanded to `entry` and `entry:*`. The SDK compares the **raw** Host header, so `example.internal` and `example.internal:8080` are different entries, while every other check in Hangar strips the port. Without this the fix reads as applied and still 421s on the port you are served on.
- `*` disables the check, matching `TrustedHostMiddleware` and `_ws_handshake_allowed`.
- Origins come from the same `MCP_CORS_ORIGINS` list the WebSocket handshake already used. A missing Origin still passes in the SDK, so non-browser clients are unaffected.

## Risk and rollback

Low, and it narrows nothing: the endpoint previously accepted only the SDK's bind-host-derived entries, and now accepts exactly what the operator configured. An unlisted host is still refused. The one behaviour change worth noting is that a deployment relying on the *accidental* `localhost:8080` entry while `MCP_TRUSTED_HOSTS` says something else will now follow the configuration — which is the intent.

Revert is a single commit.

## Notes

`trusted_hosts.py`'s module docstring already claimed `fastmcp_server.asgi` checks the MCP endpoint's Host against this list. That is true of `_ws_handshake_allowed`, which guards **WebSocket handshakes**; the HTTP MCP path had nothing. The docstring's own warning applies: *"A security allowlist that three call sites disagree about is worse than either answer consistently applied."*

## Agent metadata

- [x] LOC declared upfront: ~150

### DIFF
--- a/changelog.d/859-trusted-hosts-mcp-endpoint.fixed.md
+++ b/changelog.d/859-trusted-hosts-mcp-endpoint.fixed.md
@@ -6,4 +6,6 @@ same process accepted them -- the two read different lists. Both serving paths
 now build the guard from the configured allowlist, expanding each entry to match
 with and without a port (the SDK compares the raw `Host` header, everything else
 in Hangar strips it), with `*` opting out as it does elsewhere. Origins come from
-the same `MCP_CORS_ORIGINS` list the WebSocket handshake already used
+the same `MCP_CORS_ORIGINS` list the WebSocket handshake already used. Permitted origins are the served hosts plus `MCP_CORS_ORIGINS`, so a
+same-origin browser request keeps working while a foreign one is still
+refused
--- a/changelog.d/859-trusted-hosts-mcp-endpoint.fixed.md
+++ b/changelog.d/859-trusted-hosts-mcp-endpoint.fixed.md
@@ -1,0 +1,9 @@
+**core:** `MCP_TRUSTED_HOSTS` did not reach the MCP endpoint. The app was built
+with the SDK's default transport security, which derives its allowlist from the
+SDK's own bind host, so `/mcp` answered `421 Invalid Host header` to the
+gateway's Service DNS name and to every Ingress host while the REST API on the
+same process accepted them -- the two read different lists. Both serving paths
+now build the guard from the configured allowlist, expanding each entry to match
+with and without a port (the SDK compares the raw `Host` header, everything else
+in Hangar strips it), with `*` opting out as it does elsewhere. Origins come from
+the same `MCP_CORS_ORIGINS` list the WebSocket handshake already used
--- a/src/mcp_hangar/fastmcp_server/asgi.py
+++ b/src/mcp_hangar/fastmcp_server/asgi.py
@@ -223,6 +223,53 @@ def _ws_handshake_allowed(scope: dict) -> tuple[bool, str]:
     return True, ""
 
 
+def mcp_transport_security() -> Any:
+    """The SDK's DNS-rebinding guard, configured from Hangar's own allowlists.
+
+    The guard inside ``streamable_http_app()`` is on by default and, given no
+    settings, builds its allowlist from the SDK's default bind host. So a
+    gateway answered ``421 Invalid Host header`` to its own Service DNS name and
+    to every Ingress host, with ``MCP_TRUSTED_HOSTS`` listing them explicitly --
+    the REST API honoured that list (``TrustedHostMiddleware``) and the MCP
+    endpoint did not. On a multi-replica deployment that is the surface clients
+    use.
+
+    ``_ws_handshake_allowed`` above implements the same posture for WebSocket
+    handshakes and reads the same two lists. This is the HTTP half, which had
+    nothing.
+
+    Two translations matter:
+
+    * The SDK matches the **raw** Host header, so ``example.internal`` and
+      ``example.internal:8080`` are different entries. Hangar's own checks strip
+      the port, so each entry is expanded to both forms -- otherwise an operator
+      who wrote a hostname gets a 421 for the port they are actually served on.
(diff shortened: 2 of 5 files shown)


mapyr:
Pushed a follow-up after the conformance gate caught a regression this PR introduced, worth reading before review:

Building the transport guard's **origin** list from `MCP_CORS_ORIGINS` alone refused same-origin traffic. A browser talking to the page it came from sends `Origin: http://<served host>:<port>`, and the SDK had been deriving exactly those from its bind host — which is why the default worked and why replacing the list broke it.

| `Origin:` | first push | now |
|---|---|---|
| `http://127.0.0.1:52444` (same origin) | **403** | 200 |
| `http://localhost:52444` | **403** | 200 |
| `http://evil.example` | 403 | 403 |
| absent | 200 | 200 |

`dns-rebinding-protection` in the official suite went from passing to failing and is now back to `2 passed, 0 failed`; the whole baselined run exits 0 locally.

Origins are now the union of the served hosts (both schemes, with and without a port, IPv6 bracketed) and the configured CORS list. Unit tests cover it — the gate caught this and the tests I wrote did not, which is the more useful half of the lesson.

Issue #914, opened 2026-08-12
Approval gate can be armed with nobody listening: default delivery is a placeholder, per-policy approval_channel is never routed


mapyr:
## Summary

A deployment can configure `approval_list` on a tool policy, boot green, and have every held call sit until `approval_timeout_seconds` (default 300) and then deny — because no delivery channel ever notifies anyone. There is no boot-time refusal and no runtime signal distinguishing "gate armed and unmanned" from "gate armed and nobody has approved yet."

Verified on `mcp-hangar==2.6.0` (wheel from PyPI), not on changelog.

## Three paths to an unmanned gate

**1. The default channel is a placeholder.** `approvals/delivery/dashboard.py`:

```python
class DashboardApprovalDelivery:
    """Pushes approval notifications to connected dashboard clients via WebSocket.

    Placeholder -- actual WebSocket integration wired via event bus.
    """

    async def send(self, request: ApprovalRequest) -> None:
        logger.info("dashboard_approval_delivery", ...)
```

The docstring says the actual WebSocket integration is wired via the event bus. It is not. `approvals.channel` defaults to `dashboard` in `_build_delivery()`, and `approval_channel` defaults to `"dashboard"` in both `ToolAccessPolicy` (`domain/value_objects/tool_access_policy.py:62`) and `MCPServerConfig` (`domain/model/mcp_server_config.py:94`). The out-of-the-box gate notifies nobody.

**2. `config is None` degrades silently.** `approvals/bootstrap.py`, `_build_delivery()` returns `NoOpApprovalDelivery()` on that branch with no warning at all — unlike the unknown-channel branch below it.

**3. Unknown / unloadable / unconstructable channel degrades to noop with a boot warning only** — `approval_delivery_channel_unknown`, `approval_delivery_entry_point_failed`, `approval_delivery_construction_failed`.

The rationale in the `_build_delivery()` docstring is sound and this issue does not ask to reverse it: refusing boot over a notification channel would trade a working control plane for a dead one, and approvals do remain resolvable through the REST API. The gap is the absence of a *signal*, not the choice to keep booting.

## Why it is not fail-open, and why that is not reassuring

The failure is fail-closed by expiry: `_wait_for_decision()` times out, `ApprovalResult.expired()` returns `approved=False` with error code `approval_timeout`. Nothing leaks.

[...]

Issue #902, opened 2026-08-11
Digest pinning is inert whenever auth is off: pins are tenant-keyed and the tenant is always None


mapyr:
Verified against the published **2.5.3** wheel from PyPI (not `main`), booted with `serve --http`.

`ToolProjectionRegistry.resolve_pin` opens with:

```python
# application/read_models/tool_projection.py:263
def resolve_pin(self, mcp_server: str, tool: str, tenant_id: str | None) -> ToolDigest | None:
    if tenant_id is None:
        return None
```

and pins are declarable **only** under `tool_projection.tenant_overrides.<tenant>.pins` (`server/config.py:469-501`). There is no all-tenants form.

With auth disabled every caller is anonymous, so `tenant_id` is `None` and can never be anything else — it comes solely from `Principal.tenant_id`, and the anonymous branch hardcodes `tenant_id=None` (`fastmcp_server/asgi.py:45-54`). No header overrides it. So `resolve_pin` returns `None` for every call, `_gate_digest_pin` takes its "no pin" early return (`server/tools/batch/executor.py:1153-1155`), and the tool is invoked unchecked.

Reproduced end to end. Config with a deliberately wrong pin and the strictest mode:

```yaml
mcp_servers:
  demo:
    mode: subprocess
    command: [python, upstream.py]
    tool_projection:
      digest_enforcement: block
      tenant_overrides:
        "tenant:a":
          pins:
            echo: "0000000000000000000000000000000000000000000000000000000000000000"
```

Boot logs `auth_enabled=False` and nothing else — no warning that the pins it just parsed cannot fire. The call goes through:

```json
{"success": true, "succeeded": 1, "failed": 0,
 "results": [{"success": true, "result": {"content": [{"text": "pinned?"}]}}]}
```

Meanwhile `initialize` still advertises the capability:

```
io.mcp-hangar.digest-pinning  type=integrity  modes=[audit, warn, block]
```

Auth-off is the configuration most evaluators run. They get a declared integrity control and zero enforcement, with no signal that the two disagree.

**Second-order effect.** `set_current_tool_pin` is called only inside `_enforce_digest_pin` (`executor.py:1127`). With no pin resolved, `get_current_tool_pin()` returns `None` in `GovernedTaskStore.register_relayed_task` (`application/tasks/governed_task_store.py:213`), so relayed tasks are never bound to a digest either and the fail-closed re-verification on result retrieval is dead too. The same root cause takes out both the sync and the task path.

[...]

PR #918, opened 2026-08-12
feat(core): say when the approval gate is armed and nobody is listening

mapyr:
## Why

The substantive half of #914. **Stacked on #916** (`chore/914-retire-dashboard-channel`) — base is that branch, not `main`, because both restructure `_build_delivery`. Merge #916 first, then I rebase this onto `main`.

A deployment could put a tool on `approval_list`, boot green, and have every gated call sit for `approval_timeout_seconds` and then deny, because the delivery channel notified nobody. Nothing leaks — the gate is fail-closed by expiry — but five minutes of hanging and then an error is indistinguishable from a broken gateway, and the remediation reached for under that pressure is emptying `approval_list`. Fail-closed in code, fail-open in the organisation.

### 1. The signal

When a policy gates a tool and the channel that would notify for it reaches nothing outside the process — `noop`, or a vendor name no installed package claims — the startup check logs `subsystem_configured_but_unreachable` at ERROR, naming the scope and the channel.

It does **not** refuse the boot by default, and that is the one place I disagreed with the issue's proposal:

- The gate is already fail-closed by timeout. What is missing is a *signal*, not an enforcement — and `server/bootstrap/reachability.py` reserves `fail_closed=True` for "a gateway that cannot hold a call is a gateway executing it unapproved", which this is not.
- Refusing by default fires on the **default configuration** of every existing deployment using the gate.
- It is not the #902/#907 pattern. `a61f5ca` states the condition that made that refusal legitimate: *"the refusal has somewhere to send the operator."* Here core ships `event_stream` and `noop`, and every other channel needs a third-party package — so a refusal's only remedies would be "install someone else's adapter" or "set the escape hatch", making the escape hatch the default path.

`approvals: {delivery: {required: true}}` opts **in** to the refusal. Opt-in, not opt-out, and it reuses the existing `SubsystemRequirement` framework rather than adding a second switch beside `startup_checks.enforce`.

The check reads config, not the runtime object: `channel_reaches_a_human(channel)` resolves the same way `_build_delivery` does. That matters because the reachability tests install a `MagicMock` as `ctx.approval_gate`, and an `isinstance(gate._delivery, NoOp...)` probe would pass vacuously against a mock.

### 2. `approval_channel` finally routes

It was documented as a policy's delivery channel and merged with care across scope narrowing — `min` timeout, channel taken from the narrower policy that owns the `approval_list`, in seven branches — and dispatched nowhere. One global delivery took every approval whichever policy raised it, so two servers with different channels silently shared one.

`ApprovalDeliveryRouter` satisfies the `ApprovalDelivery` protocol itself, so `ApprovalGateService` calls `send` exactly as before and learns nothing about channels. Channels build on first use rather than being enumerated at boot, because a policy can arrive later from a hot reload or over REST.

**Semantic change worth flagging:** `ToolAccessPolicy.approval_channel` now defaults to **empty**, meaning "the deployment's `approvals.channel`". With a concrete default every policy would route by that name and leave the global setting nearly dead. The service resolves the effective channel once, so the record, the event, the span attribute and the metric all say the same thing the router will do. (This supersedes the policy-default half of #916; the global `DEFAULT_CHANNEL` there is unchanged.)

### 3. Metrics

Approvals had none. `mcp_hangar_approval_requests{channel}` against `mcp_hangar_approval_deliveries{channel,outcome}` (`sent` / `failed` / `not_notified`) and `mcp_hangar_approval_decisions{channel,decision}` (`granted` / `denied` / `expired`). Requests climbing while deliveries stay at zero is the armed-and-unmanned shape; `expired` climbing beside a flat `sent` is the same story from the other end.

### Not done here

`send() -> None` stays — the notifier must not decide. Runtime degrade-to-noop for a missing adapter is unchanged. Both are the issue's stated non-goals. Docs (`APPROVAL_ADAPTERS.md`, `configuration.md`, ADR-016) follow in the docs repo once the channel names are settled.

## How tested

`tests/unit/test_armed_approval_gate_says_if_it_is_unmanned.py` — 19 tests over which channels reach anyone, the check's default and opted-in behaviour, per-policy channel selection, and routing including a channel first seen after boot.

**Verified the tests catch the bug**: with `_approval_delivery_requirements` unregistered, 5 of them fail. Restored, all 19 pass.

Full unit suite green (6620 passed, 2 skipped); integration green (259 passed, 27 skipped).

Refs #914


mapyr:
Reopened as #920. This PR shows `MERGED` but none of its content reached `main`: it was based on #916's branch, and squash-merging #916 closed this one without landing its commits.

```console
$ git show origin/main:src/mcp_hangar/metrics.py | grep -c mcp_hangar_approval
0
$ git show origin/main:src/mcp_hangar/server/bootstrap/reachability.py | grep -c _approval_delivery_requirements
0
```

Same three commits, cherry-picked onto merged `main`, no content change.

PR #923, opened 2026-08-13
fix(core): import flat projection before server bootstrap

### LINKED ISSUE #894: flat_tool_projection cannot be imported first: a circular import the suite hides
`flat_tool_projection` cannot be imported first. On clean `main` @ `c4315a1`:

```console
$ python -c "import mcp_hangar.fastmcp_server.flat_tool_projection"
  File ".../server/bootstrap/__init__.py", line 38, in <module>
    from ...fastmcp_server.flat_tool_projection import maybe_register_flat_tool_handlers
ImportError: cannot import name 'maybe_register_flat_tool_handlers' from partially
initialized module 'mcp_hangar.fastmcp_server.flat_tool_projection'
(most likely due to a circular import)
```

The cycle: `flat_tool_projection` imports `..server.tools.batch` (l. 59) → that package chain reaches `server.bootstrap` → which imports `flat_tool_projection` back (l. 38) while it is still initialising.

## Why it has gone unnoticed

Nothing in production imports it first — `bootstrap` always wins the race — and the test suite hides it because some earlier test always imports `mcp_hangar.server` before any test touches the projection module. It only shows when a single test file is run on its own:

```console
$ uv run pytest tests/unit/test_tenant_scoped_cachescope.py
ERROR tests/unit/test_tenant_scoped_cachescope.py
!!!!!! Interrupted: 1 error during collection !!!!!!
```

That file is on `main` today and passes in a full-suite run. So the suite is order-dependent in a way nobody chose, and `pytest <one file>` — the thing anyone does while iterating — fails with an import error that has nothing to do with the file.

## Why it is worth fixing rather than noting

Two reasons beyond the papercut:

1. **It is a layering violation with a working detector attached.** `import-linter` is already in CI and did not catch this, so whatever contract covers `fastmcp_server` does not forbid the reach into `server.tools.batch`. Worth finding out why, since that is the check that should have.
2. **The direction is wrong.** `flat_tool_projection` is a delivery-layer module reaching into `server.tools.batch` for `BatchExecutor` / `CallSpec` at module scope, purely to dispatch a flat call. A deferred import inside `_flat_call_tool`, or moving the executor behind a port, both break the cycle — the second is the one that matches how the rest of the package is wired.

## Found while

Writing tests for #880, whose new test file has to carry an explicit `import mcp_hangar.server` with a comment explaining why. That workaround should be deleted when this is fixed.

Liona-orph:
## Why

Importing `mcp_hangar.fastmcp_server.flat_tool_projection` as the first Hangar module raised an ImportError because the projection imported server tooling before the composition root had finished initializing. This prevents isolated tooling, tests, and embedders from importing the public projection module directly. Closes #894

## What

- Defer the batch executor and management permission imports until request handlers run.
- Add a fresh-process regression test that imports the projection before `mcp_hangar.server`.
- Add the required changelog fragment.

## How tested

- `uv run pytest tests/unit/test_flat_tool_projection_import.py -q`
- `uvx ruff@0.16.1 check src/mcp_hangar/fastmcp_server/flat_tool_projection.py tests/unit/test_flat_tool_projection_import.py`
- `uv run python scripts/build_changelog.py check`

## Risk and rollback

Low risk; request-time imports preserve the existing runtime dependencies and behavior. Revert the single commit to restore the prior import layout.

## CHANGELOG note

`changelog.d/894-flat-projection-import.fixed.md`: **core:** the flat tool projection can now be imported as the first Hangar module; its batch executor dependency is loaded when a call is dispatched, after bootstrap has finished wiring the serving surface

## Agent metadata

- [x] LOC declared upfront: 24 (excluding tests)


mapyr (commented):
Thanks for this — the diagnosis is right and the test is the right shape. Reproduced the bug on `main` before reading the patch:

```console
$ python -c "import mcp_hangar.fastmcp_server.flat_tool_projection"
ImportError: cannot import name 'maybe_register_flat_tool_handlers' from partially
initialized module 'mcp_hangar.fastmcp_server.flat_tool_projection'
(most likely due to a circular import)
```

and with the patch applied: `imported OK`. Spawning a fresh interpreter in the test is exactly right — an in-process test cannot see this, because by then something has already imported the server package.

## One thing to fix: three existing tests break

CI has it, and it is not a flake:

```
FAILED tests/unit/test_flat_tool_reexport.py::TestFlatCallToolHandler::test_valid_call_routes_through_enforcement_and_returns_result
FAILED ...::test_toctou_tool_withdrawn_between_list_and_call
FAILED ...::test_enforcement_failure_surfaces_as_tool_error
  AttributeError: module 'mcp_hangar.fastmcp_server.flat_tool_projection'
                  does not have the attribute 'BatchExecutor'
```

Those tests patch the re-exported name at lines 622, 714 and 775:

```python
patch("mcp_hangar.fastmcp_server.flat_tool_projection.BatchExecutor", ...)
```

Moving the import inside the function removes that module attribute, so there is nothing left to patch. **Repoint them at the definition site** — the function-local `from ..server.tools.batch import BatchExecutor` reads the attribute off the source module at call time, so patching there still intercepts it:

```python
patch("mcp_hangar.server.tools.batch.BatchExecutor", ...)
```

Verified rather than suggested: with only those three strings changed, that file goes from `3 failed, 17 passed` to **`20 passed`**.

Also `ruff format` wants a blank line after `from mcp_hangar._sdk_compat import CallToolResult` — that is the whole of the `lint` failure.

## Worth knowing, since it makes the case for this PR stronger

`tests/unit/test_flat_tool_reexport.py` **cannot be collected in isolation on `main` at all** — `pytest tests/unit/test_flat_tool_reexport.py` dies on this same circular import. It only passes today because a full-suite run happens to import the server package first. So #894 is not just an ergonomic problem for a fresh interpreter; it already makes part of the test suite order-dependent, and this PR fixes that too. Might be worth a line in the changelog fragment.

Nothing else from me — the lazy-import comments explain *why* at both sites, which is the part that stops someone hoisting them back later.

mapyr:
Pushed a follow-up commit; the three red tests were not a problem with the change.

Deferring the import removed the module-level `BatchExecutor` name, and `test_flat_tool_reexport.py` patched `flat_tool_projection.BatchExecutor` in three places. Repointed at `mcp_hangar.server.tools.batch.BatchExecutor`, which is where the handler's deferred import resolves it — there is no module-level name to patch any more, and there must not be one.

Also in this commit:

- deleted the two import-order workarounds the cycle forced on test files (`test_empty_front_door_is_explained.py`, `test_front_door_sees_the_caller_tenant.py`) — #894 explicitly asks for that, and all three formerly order-dependent files now pass run on their own;
- dropped the `if TYPE_CHECKING: pass` block the moved imports left behind;
- `ruff format` on the new test file (the red `lint` job was `ruff format --check src/mcp_hangar` failing on the comment placement in `flat_tool_projection.py`, fixed too);
- simplified the fresh-process test: the package is installed, so the subprocess resolves it without the PYTHONPATH/cwd setup.

Local: full `tests/unit` green (6637 passed), `ruff format --check src/mcp_hangar` clean.

**Point 1 of the issue is not addressed here and needs a decision.** The issue asks why `import-linter` did not catch this. Answer: it cannot, as configured. The `hexagon` contract puts `mcp_hangar.server : mcp_hangar.fastmcp_server : mcp_hangar.facade` on one layer line, and nothing forbids edges *within* a layer, so `fastmcp_server -> server.tools.batch` is invisible to it — `lint-imports` reports `Contracts: 1 kept, 0 broken` on the broken main too. Making them independent would flag the deferred imports as well (grimp counts function-scope imports), so it costs an `ignore_imports` entry, and `tests/unit/test_import_contracts.py` caps that ledger so it cannot grow. That is a deliberate call about the ledger, not a fix to slip into this PR.

PR #924, opened 2026-08-13
fix(core): reject unknown discovery modes

### LINKED ISSUE #832: An unknown discovery `mode` resolves to additive silently, so a misspelt `authoritative` stops removing servers
`create_source` (`infrastructure/discovery/registry.py:110-111`) resolves the configured mode like this:

```python
mode_str = str(config.get("mode", "additive"))
mode = DiscoveryMode.AUTHORITATIVE if mode_str == "authoritative" else DiscoveryMode.ADDITIVE
```

Anything that is not exactly `authoritative` becomes ADDITIVE, silently. The dangerous direction is a typo in the *authoritative* spelling: an operator writes `authoritativee`, or `Authoritative`, and gets a source that only ever adds. Nothing removes a server that has gone away, no warning is logged, and `GET /api/discovery/sources` reports `mode: additive` -- which is true, and is not what was asked for.

```
create_source('docker', mode='additivee')      -> mode = additive   (harmless)
create_source('kubernetes', mode='Authoritative') -> mode = additive   (silently wrong)
```

An unknown `source_type` is already refused outright, and `UPGRADE.md` explains why: a configured source that silently does nothing is the failure this codebase keeps finding. A mode is the same kind of statement about what a source is for.

## Proposed fix

Parse it with `DiscoveryMode(mode_str)` and let the `ValueError` name the value and the two valid spellings, the way `init_event_store` refuses an unknown driver. This is a startup refusal on a configuration that was already not doing what it said, so it belongs in `UPGRADE.md` alongside the unknown-`source_type` note.

Split out of #828, which fixed the registry/orchestrator divergence that this tolerance exposed. Fixing that one does not fix this one: after #831 a misspelt mode is consistently visible and scannable, and still running in the wrong mode.

Liona-orph:
## Why
Closes #832. A misspelled authoritative mode was silently downgraded to additive, so a source could stop removing servers while the gateway appeared healthy.

## What
Parse configured values through `DiscoveryMode` and raise an actionable `ValueError` for unknown values. Add unit regressions for misspellings and case variants, plus an upgrade note and changelog fragment.

## How tested
- `uv run pytest tests/unit/test_discovery_source_registry.py -q`
- `uvx ruff@0.16.1 check src/mcp_hangar/infrastructure/discovery/registry.py tests/unit/test_discovery_source_registry.py`
- `uv run python scripts/build_changelog.py check`

## Risk and rollback
This changes only invalid discovery configuration; omitted, `additive`, and `authoritative` values keep their existing behavior. Roll back the PR commit if a deployment depends on the old silent fallback.

## CHANGELOG note
`changelog.d/832-reject-unknown-discovery-mode.fixed.md` documents the startup validation and upgrade action.

## Agent metadata
- Implemented and tested by Liona-orph
- Scope limited to issue #832
- No generated files or dependency changes


mapyr (commented):
Thanks — the direction is right, and shipping it with an `UPGRADE.md` entry and a changelog fragment is the correct instinct for a behaviour change. Two things to resolve, one of which changes what the docs should say.

## 1. The six failing tests are not a wrong change — they are the wrong home

```
FAILED tests/unit/test_the_registry_and_the_orchestrator_agree.py::TestAMisspeltModeIsStillScannable
  test_the_builder_accepts_it
  test_the_source_the_listing_advertises_can_be_scanned
  test_the_spec_carries_the_mode_the_source_is_running_in
  test_whatever_was_written_the_two_never_disagree[additivee]
  test_whatever_was_written_the_two_never_disagree[AUTHORITATIVE]
  test_whatever_was_written_the_two_never_disagree[]
```

That class is not endorsing the lenient behaviour, and says so:

> Stated as a fact this test depends on, not as an endorsement: an unknown mode resolves to additive rather than being refused.

The invariant the file actually guards is *the registry lists what the orchestrator runs*. A misspelt mode was one vehicle for a real past defect — the listing handed out an id and `POST /api/discovery/sources/<id>/scan` answered 404 for it.

Under your change that case simply moves house. `create_source` raises, so the orchestrator has no source; the second reading already did `DiscoveryMode(mode_str)`, caught `ValueError` and skipped. Both readings now say nothing, so they agree — which is the same shape as `TestASourceThatDidNotBuildIsNotAddressable` further down the file.

So the fix is to move `additivee`, `AUTHORITATIVE` and `""` out of `TestAMisspeltModeIsStillScannable` into the did-not-build class, and update the module docstring's "**A misspelt mode**" paragraph, which describes behaviour this PR removes. The invariant survives intact; only the example changes.

## 2. `UPGRADE.md` promises something the code does not do

Both the changelog fragment and the upgrade note say startup refuses. It does not. `server/bootstrap/discovery.py` catches this at the call site:

```python
except UnknownDiscoverySourceError:   # deliberately re-raised
except ImportError as e:              # degrade
except Exception as e:                # <- ValueError lands here: log and continue
```

Confirmed by running it: `create_source("filesystem", {"mode": "Authoritative", ...})` raises your `ValueError`, and bootstrap logs `discovery_source_error` and carries on **without that source**.

The comment above the re-raise branch is worth reading, because it describes exactly the outcome this leaves:

> A configured source that silently does nothing is the failure this codebase keeps finding: the operator believes the fleet is watched and one startup warning is the only thing that ever says otherwise.

For a misspelt `authoritative` the before/after becomes: *runs as additive and never removes* → *does not exist at all and never discovers*. Better, in that the listing no longer advertises a source that cannot be scanned — but not "refuses to start".

Two ways out, either is fine:

- **Make the promise true** — raise `UnknownDiscoverySourceError` (or a sibling added to that re-raise list). It is the same class of mistake the existing branch already refuses to swallow, so this reads as consistent rather than special-cased, and #832 is arguably asking for exactly that.
- **Make the docs true** — keep `ValueError` and say the source is dropped with a `discovery_source_error` log line rather than that startup refuses.

I would take the first: a deployment that names an `authoritative` source and silently gets none is the failure mode #832 opened about.

Also `ruff format` collapses the new `raise ValueError(...)` onto one line — that is the whole `lint` failure.

Nothing wrong with the core change itself; `DiscoveryMode(mode_str)` with the valid values in the message is the right shape, and catching capitalisation variants is a good call.

mapyr:
Pushed a follow-up commit: the refusal this PR added did not actually refuse.

`create_discovery_orchestrator` wraps source construction in a fault barrier that catches every `Exception` and logs `discovery_source_error`, so the bare `ValueError` was swallowed there. A misspelt mode went from *silently additive* to *silently absent* — the same failure this issue is about, relocated. The refusal now travels as a named `UnknownDiscoveryModeError`, re-raised beside `UnknownDiscoverySourceError` in the clause that already exists for exactly this reason, and there is a bootstrap-level test that fails if it stops escaping the barrier.

Also rewrote `TestAMisspeltModeIsStillScannable` in `tests/unit/test_the_registry_and_the_orchestrator_agree.py`, which is what the 6 red tests were: that class comes from #831 and asserts the tolerance this PR removes. The invariant it exists for — the registry listing and the orchestrator never disagree — is kept, now over the modes that do build. Its docstring records why the divergence closed.

Local: full `tests/unit` green (6638 passed), `ruff format --check src/mcp_hangar` clean.

One thing deliberately left out of scope: `POST /api/discovery/sources` reaches `DiscoveryMode(command.mode)` in `application/commands/discovery_handlers.py:60` and raises a bare `ValueError` with no 4xx mapping, so a misspelt mode over REST is likely a 500. Same family, different surface — worth its own issue rather than widening this one.

PR #925, opened 2026-08-13
fix(ci): isolate project add runs

### LINKED ISSUE #847: ci: Project Add cancels itself across PRs, so a batch never reaches the board
## What

`.github/workflows/project-add.yml` keys its concurrency group on the workflow name and `github.ref`:

```yaml
on:
  pull_request_target:
    types: [opened, reopened, ready_for_review]
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

Under `pull_request_target` the workflow runs in the **base branch** context, so `github.ref` is
`refs/heads/main` for every pull request. All PRs therefore share one concurrency group with
`cancel-in-progress: true`, and opening two close together cancels the earlier one.

## Evidence

Opening three PRs within seconds today: the last one succeeded, the two before it were **cancelled**.

| run | head branch | conclusion |
| --- | --- | --- |
| 31407053070 | `docs/hangar-sources-docstring-carries-the-id` | success |
| 31407050205 | `fix/coordination-without-a-selected-backend` | cancelled |
| 31407044937 | `security/the-connect-time-guard-does-not-survive-a-restart` | cancelled |

Re-running the two cancelled runs one at a time made both succeed with no change to the PRs, which
is the confirmation that the content was never the problem.

Every older run in the list is green, because those PRs were opened one at a time.

## Why it matters more than the red check

`add` is not a required check, so nothing is blocked. What actually happens is that the affected
PRs are **never added to the project board** — silently, and specifically when a batch of related
work is opened together, which is exactly when the board is most useful.

In one session this hit **6 PRs across 2 repositories**: this repo, and `mcp-hangar/docs`, whose
`project-add.yml` carries the same group expression (4 of 5 PRs cancelled there).

## Fix

Key the group on the pull request instead of the base ref:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.issue.number || github.ref }}
  cancel-in-progress: true
```

The `issue.number` fallback keeps the `issues:` trigger grouped per issue, and the `github.ref`
fallback keeps any other trigger behaving as before.

Worth applying in `mcp-hangar/docs` at the same time, and worth a grep across the org for
`github.workflow }}-${{ github.ref` on other `pull_request_target` workflows — the expression is
correct under `pull_request` and wrong here, so it copies between files without looking wrong.

Liona-orph:
## Why
Closes #847. `pull_request_target` evaluates `github.ref` against the base branch, so concurrent pull requests shared one group and cancelled one another before they could be added to the project.

## What
Key the concurrency group by pull request number, with issue and ref fallbacks for the other supported trigger. Add a regression test that guards the event-specific expression and cancellation setting.

## How tested
- `uv run pytest tests/ci/test_project_add_workflow.py -q`
- `git diff --check`

## Risk and rollback
Only concurrency grouping changes. Runs for the same pull request or issue still replace one another, while separate pull requests no longer cancel each other. Revert the workflow commit to roll back.

## CHANGELOG note
`skip-changelog: CI-only workflow concurrency fix`

## Agent metadata
- Implemented and tested by Liona-orph
- Scope limited to issue #847
- No application runtime changes

### DIFF
--- a/.github/workflows/project-add.yml
+++ b/.github/workflows/project-add.yml
@@ -7,7 +7,7 @@ on:
     types: [opened, reopened, ready_for_review]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.issue.number || github.ref }}
   cancel-in-progress: true
 
 permissions:
--- a/tests/ci/test_project_add_workflow.py
+++ b/tests/ci/test_project_add_workflow.py
@@ -1,0 +1,24 @@
+"""Regression tests for the project-add workflow's event-specific concurrency."""
+
+from pathlib import Path
+import re
+
+
+WORKFLOW = Path(__file__).parents[2] / ".github" / "workflows" / "project-add.yml"
+
+
+def _concurrency_group_expression(workflow: str) -> str:
+    match = re.search(r"^\s+group:\s+\$\{\{\s*(.*?)\s*\}\}\s*$", workflow, re.MULTILINE)
+    assert match is not None, "project-add must define a concurrency group expression"
+    return match.group(1)
+
+
+def test_project_add_keeps_pull_request_runs_in_separate_groups() -> None:
+    workflow = WORKFLOW.read_text()
+    expression = _concurrency_group_expression(workflow)
+
+    assert "pull_request_target:" in workflow
+    assert "github.event.pull_request.number" in expression
+    assert "github.event.issue.number" in expression
+    assert "github.ref" in expression
+    assert "cancel-in-progress: true" in workflow


mapyr (commented):
Looks good to me — no changes requested. 26/26 green.

Checked the parts that are easy to get wrong in a concurrency expression:

- both triggers are covered. `pull_request_target` populates `github.event.pull_request.number`; `issues` populates `github.event.issue.number`; the `||` chain picks the right one for each and still falls back to `github.ref`, so nothing loses a group.
- `cancel-in-progress: true` is now correct rather than harmful: with the group keyed per PR, a superseding run only cancels the same PR's earlier run, which is what that setting is for. Keyed on `github.ref` it was cancelling *other* PRs, which is #847.
- no permission change, no new step, no checkout of fork code.

Worth noting for anyone reviewing this from a fork-safety angle: this edits a `pull_request_target` workflow, but the version that executes comes from the base branch, so the change cannot run itself here. It takes effect after merge.

On the test — asserting against the YAML text is weak, and normally I would push back. Here it is the right call: there is no way to exercise a workflow's `concurrency` semantics without running Actions, and pinning the expression's operands is a real guard against someone simplifying it back to `github.ref`. `tests/ci/` is a new directory but `testpaths = ["tests"]` picks it up, so it runs.

One optional nit, take it or leave it: `assert "pull_request_target:" in workflow` checks something this PR does not change, so it will keep passing even if the concurrency fix is reverted. The three `in expression` assertions are the ones carrying the weight.

Issue #1059, opened 2026-08-24
[BUG] the three approval-gate metrics are never registered, so /metrics never carries them


mapyr:
## Short summary

The three approval-gate counters are defined, incremented on the live code path,
and documented with recommended PromQL — and never appear on `/metrics`.
`_register_all_metrics()` (`src/mcp_hangar/metrics.py`) never adds them to
`REGISTRY`, and `CollectorRegistry.collect()` only walks registered collectors.
Every one of the three queries in `guides/OBSERVABILITY.md` returns an empty
result on any deployment.

Same shape as `PROJECTED_TOOLS` / `EMPTY_PROJECTION_TOTAL`, which are missing
from the same list and are being wired up alongside #1049 / #1053. These three
are the remainder.

## Reproduction steps

```python
import mcp_hangar.metrics as m

m.APPROVAL_REQUESTS_TOTAL.inc(channel="slack")
print("mcp_hangar_approval_requests" in m.get_metrics())   # False
```

Or statically, against `main`:

```python
import re
src = open("src/mcp_hangar/metrics.py").read()
defined = set(re.findall(r'^([A-Z][A-Z0-9_]+)\s*=\s*(?:Counter|Gauge|Histogram)\(', src, re.M))
registered = set(re.findall(r'\b([A-Z][A-Z0-9_]+)\b', src.split("def _register_all_metrics")[1]))
print(sorted(defined - registered))
# ['APPROVAL_DECISIONS_TOTAL', 'APPROVAL_DELIVERIES_TOTAL', 'APPROVAL_REQUESTS_TOTAL',
#  'EMPTY_PROJECTION_TOTAL', 'PROJECTED_TOOLS']
```

Over the wire: arm the approval gate, hold one tool call, decide it, then
`GET /metrics`. No `mcp_hangar_approval_*` series is present — not zero-valued,
absent.

## Expected behavior

An operator can answer the three questions `guides/OBSERVABILITY.md` tells them
to ask:

```promql
sum(rate(mcp_hangar_approval_requests_total[15m])) by (channel)
  - sum(rate(mcp_hangar_approval_deliveries_total{outcome="sent"}[15m])) by (channel)

sum(rate(mcp_hangar_approval_decisions_total{decision="expired"}[1h])) by (channel)
  / sum(rate(mcp_hangar_approval_decisions_total[1h])) by (channel)

sum(rate(mcp_hangar_approval_deliveries_total{outcome="failed"}[5m])) by (channel)
```

[...]

Issue #1056, opened 2026-08-24
[FEAT] withdraw a tool whose `x-mcp-header` annotations are invalid, at projection time


mapyr:
## Problem statement

A modern client **MUST** drop any tool whose `x-mcp-header` annotations fail
`find_invalid_x_mcp_header` (`mcp/client/session.py:1253-1266`). Hangar forwards
the whole upstream `inputSchema` (`flat_tool_projection.py`, `payload = dict(proj.schema)`
at the `_build_mcp_tool_list` loop) and counts the tool in `PROJECTED_TOOLS`.
Projected ≠ usable, with no server-side signal.

Part of #1054.

## Proposed solution

Run `find_invalid_x_mcp_header` when the tool enters the projection (or when
the flat list is built — one place, not both). An invalid annotation **withdraws**
the tool: named reason, log line, metric. Do not strip the annotation in place;
that mutates a JCS digest over `{name, description, inputSchema, outputSchema}`.

## Alternatives considered

Strip the annotation — rejected, breaks pinning. Deny the whole upstream — too
blunt; the SDK's own rule is per-tool.

## Acceptance criteria

* An upstream tool with an invalid `x-mcp-header` is absent from a front-door
  `tools/list` and is not counted by `mcp_hangar_projected_tools`.
* A log line names the tool and the reason; a metric counts the withdrawal.
* The digest of a *served* tool is byte-identical to the upstream's.
* **Verified by** driving `tools/list` through the front door against an upstream
  stub whose schema carries (a) a valid annotation, (b) an invalid one — not by
  unit-testing the SDK validator.

## Scope

security

Issue #1053, opened 2026-08-24
[BUG] `Mcp-Param-*` validation skips are unobserved, and some skip the check while the call still proceeds


mapyr:
## Short summary

SEP-2243's header–body agreement is what stops a caller from routing on one value
while executing another. The SDK enforces it for `Mcp-Param-*`, but the
enforcement is fail-open by design: if the pre-dispatch `tools/list` cannot
produce a schema, validation is skipped and the call proceeds.

**Rewritten after 2.14.0.** The original text said "Hangar does not currently
route or authorize on `Mcp-Param-*`", so the worst case was a deployment that put
an intermediary in front of us. That premise died in the same release: #1064
gives `MCPEgressPolicy` a `headers.allow` / `deny` / `requireApproval` ladder that
selects on `Mcp-Param-*`, and `context.py:routing_headers_var` carries those
headers to the egress evaluator and the batch executor. We *are* the intermediary
now, and the "MUST NOT treat as trusted" obligation is no longer met by
construction -- it has to be shown.

The boundary this issue names is therefore not the metric, and not the skip in
the abstract. It is that **an L7 header selector trusts a header nobody compared
against the body**.

## The boundary

`evaluate_headers` (`domain/policies/egress_l7.py:356`) gates on the protocol
version alone:

```python
if headers is None or not is_modern_protocol_version(headers.get(PROTOCOL_VERSION_HEADER)):
    return None
```

The docstring on `context.py:47` says the version "says whether they were checked
against the body at all". It does not. It says whether they *should have been*.
On the `listing_failed` arm -- a modern request where the SDK's pre-dispatch
listing raised, the SDK catches at `_streamable_http_modern.py:251-256`, returns
`None` and lets dispatch continue -- the selector matches `allow` / `deny` /
`requireApproval` against an unvalidated header. That is the SEP's "route on one
value, execute another", inside Hangar rather than in front of it. The 2.14.0
changelog sentence ("never satisfies such a selector") covers only the
handshake era.

## Skip is not gap: per reason

The four metric reasons are not one class, and any fail-closed switch hits only
one of them:

* `tool_not_listed` -- dispatch answers `-32601`. Not a gap, given the
  same-principal assertion in #1049.
* `invalid_annotation` (`flat_tool_projection.py:526`) -- since #1063 such a tool
  is withheld from the projection at `flat_tool_projection.py:293`, so the lookup
  misses and it is counted as `tool_not_listed`. Reachable only for a `hangar_*`
[...]

PR #1060, opened 2026-08-24
fix(core): decode SEP-2243 routing headers, answer -32020, and memoise the projection

### LINKED ISSUE #1049: [BUG] a modern-protocol `tools/call` re-runs the whole front-door projection, and the governance-cost metrics count it
## Short summary

To validate SEP-2243 `Mcp-Param-*` headers, the SDK's modern streamable-HTTP
transport resolves the called tool's `inputSchema` by **invoking our own
`tools/list` handler**, pre-dispatch, on every `tools/call` that carries
arguments. There is no cache. For the front door that means the identity-scoped
projection is rebuilt, and `PROJECTED_TOOLS` — shipped in #912 as the first
measurement of the governance surface's cost in a client's context budget —
records an observation per **call** instead of per **listing**. The metric now
measures traffic.

The call handler also rebuilt `_build_flat_map` independently (`flat_tool_projection.py`
after the list), so an argument-bearing modern `tools/call` paid for the map
twice even before counting the SDK listing.

## Reproduction steps

Run a front-door deployment, negotiate `2026-07-28`, list once, then issue a
single `tools/call` **with arguments** and scrape `/metrics`.

`PROJECTED_TOOLS.observe()` fires twice per listing (`kind=governed` and
`kind=management`), so `mcp_hangar_projected_tools_count` advances by **2** for
a client-visible `tools/list`, not 1. After a `tools/call` with arguments it
advances again, with no client-visible listing in between.

`EMPTY_PROJECTION_TOTAL` follows the same path (`_report_empty_projection` at
`flat_tool_projection.py:377+`) and is subject to the same inflation; its 1/min
log throttle protects the log, not the counter.

Source path, read on the installed artifacts (`mcp-hangar 2.13.0`, `mcp 2.0.0`):

* `mcp/server/_streamable_http_modern.py:273-303` — `_mcp_param_rejection` returns
  early only when the call has **no arguments and no `Mcp-Param-*` headers**;
  otherwise it resolves the schema.
* `:205-270` — `_tool_input_schema` calls `serve_one(app, …, "tools/list", …)`
  through the normal path, a fresh `Connection` per page, up to
  `_MCP_PARAM_LIST_PAGE_CAP` pages. The docstring is explicit that this is
  deliberate: *"a visibility-scoped catalog yields exactly what this caller was
  advertised."*
* `mcp_hangar/fastmcp_server/flat_tool_projection.py` — `PROJECTED_TOOLS.observe()`
  sits inside `_flat_list_tools`, so it fires for the pre-dispatch listing too.

## Expected behavior

`mcp_hangar_projected_tools` describes the surface a client was handed, once per
listing that client actually received (one pair: governed + management). A
`tools/call` does not emit a new pair. The pre-dispatch schema lookup is served
from work the request already did (or will do): the map is built at most once
per HTTP request.

## Actual behavior

Every argument-bearing `tools/call` on a modern-protocol connection pays a second
full projection — identity resolution, policy resolver, `_build_flat_map`,
`_build_mcp_tool_list` — and emits a second pair of histogram observations. The
measured local overhead (p50 +3.5 ms / p95 +9.0 ms) predates this path and does
not include it.

Identity on the pre-dispatch listing: it runs through the registered `_list_v2`
adapter, which calls `bind_caller_identity` (the #856 / #874 bridge) on the same
Starlette `Request` as the call. That is the same principal, not an unauthenticated
projection — but nothing asserted it until this issue. If the bridge were skipped,
the listing would be `EMPTY_NO_IDENTITY` and `Mcp-Param` validation would be skipped
(#1053). The acceptance test below is that assertion.

## Version

2.13.0. Introduced by the move to the SDK's modern transport, not by any Hangar
change — which is why no changelog entry marks it.

## Severity

p2-normal. No enforcement gap on the identity path as wired; a cost and a
correctness defect in the one metric we point at when arguing that a governance
plane is not free. A metric that silently changes meaning is worse than an
absent one, because it is quoted.

## Fix

1. **Serve the pre-dispatch listing from a per-request memo** on `request.state`,
   keyed on tenant. The call handler reuses the same map. This also removes the
   double policy evaluation, which is the part that scales with catalog size.
2. **Observe the metric only for a client-visible listing.** Detect the HTTP
   envelope method from the cached request body (`tools/call` → do not observe).
   Leaving the observation inside the builder without that gate means any future
   internal caller re-inflates it.

Not a patch to the SDK: the behavior is spec-mandated and documented.

## Acceptance

* One `tools/call` with arguments produces **zero** new `mcp_hangar_projected_tools`
  observations and zero `empty_projection_total` increments.
* A listing the client actually receives still produces exactly one pair
  (governed + management).
* The projection is built at most once per HTTP request; asserted by a counter or
  a spy on the builder, not by timing.
* A test asserts the pre-dispatch listing resolves the **same principal** as the
  call it is validating (the #874 class), and fails loudly if it resolves to
  anonymous.

Related: #1053, #1054.

### LINKED ISSUE #1050: [BUG] a base64-encoded `Mcp-Name` mismatches on the handshake-era front door and on `tasks/*`
## Short summary

SEP-2243 requires a client to wrap an `Mcp-Name` value in the `=?base64?…?=`
sentinel whenever the value is non-ASCII, has leading/trailing whitespace, or
carries control characters. `Mcp-Name` for `resources/read` is the **URI**, so a
non-ASCII resource path is the ordinary case, not the exotic one.

Hangar's handshake-era middleware compares the header to the body as raw latin-1
text and never decodes the sentinel. A 2026-07-28 client does **not** hit this
path: `FrontDoorRoutingMiddleware._should_engage` returns False and the SDK
inbound ladder already calls `decode_header_value` (`mcp/shared/inbound.py:447`).
The production blast radius is handshake-era POSTs that carry routing headers,
plus `tasks/*` (`_require_mcp_name_header`), which the SDK does not cover.

`encode_header_value` leaves plain ASCII without edge whitespace verbatim, so a
typical ASCII `taskId` is unaffected. Non-ASCII or whitespace-padded values wrap
and were refused.

## Reproduction steps

Handshake-era (the middleware is the owner):

```python
from mcp.shared.inbound import encode_header_value

uri = "file:///Załączniki/raport.pdf"
# POST /mcp with MCP-Protocol-Version: 2025-06-18
# Mcp-Method: resources/read
# Mcp-Name: encode_header_value(uri)
# body method resources/read, params.uri = uri
```

`resolve_route` on the raw sentinel also raises `HeaderBodyMismatchError`; that
helper is not what production executes on 2026-07-28.

## Expected behavior

The front door decodes the sentinel before comparing, exactly as the SDK's
inbound ladder does, so the two paths in this product agree on what "the header
matches the body" means. A base64-wrapped value that decodes to the body value
routes; one that decodes to something else is still a mismatch. `tasks/*` uses
the same codec.

## Actual behavior

* `extract_route_headers` (`fastmcp_server/front_door_routing.py`) did
  `value.decode("latin-1").strip()` and nothing else.
* `resolve_route` compares `header_name != body_name` as strings (fine once
  extract has decoded).
* `_require_mcp_name_header` (`fastmcp_server/task_relay_handlers.py`) had the
  same raw comparison against the body `taskId`.

The failure is fail-closed, so nothing leaks. It refuses correct handshake-era
and `tasks/*` traffic that used the sentinel.

## Version

2.13.0. Present since the middleware / task gate were introduced.

## Severity

p2-normal. No enforcement gap. A 400 on a Polish `resources/read` URI looks like
a Hangar outage **on handshake-era clients that send SEP-2243 headers**. Modern
clients are already decoded by the SDK.

## Fix

Route both comparisons through `mcp.shared.inbound.decode_header_value`. Decode
in `extract_route_headers` so `RouteDecision.name` carries the real value
downstream; decode in `_require_mcp_name_header` before the `taskId` comparison.
A malformed sentinel must not be treated as "header absent" (that would fall
through to the body).

## Acceptance

* A handshake-era `resources/read` whose `uri` is non-ASCII, sent with a
  sentinel-wrapped `Mcp-Name`, routes and is **not** refused. **Verified by**
  driving the middleware over a real ASGI request.
* A sentinel-wrapped `Mcp-Name` that decodes to a *different* body value is still
  refused, with `HEADER_MISMATCH` (#1051).
* A value with leading/trailing whitespace (the spec's other base64 trigger) is
  covered by the same ASGI test.
* `tasks/*` accepts a base64-wrapped `taskId` that decodes to the body, and
  refuses one that does not.

Related: #1051.

### LINKED ISSUE #1051: [BUG] a front-door header/body mismatch is refused as -32600, not `HeaderMismatch` (-32020)
## Short summary

SEP-2243 (Final) requires a header/body mismatch to be answered with HTTP 400 and
the `HeaderMismatch` JSON-RPC code. This package already defines that code —
`tasks_wire.HEADER_MISMATCH = -32020`, carrying the errata from
modelcontextprotocol#2907 that moved it off `-32001` — and the task relay uses it.
The handshake-era front door, refusing the same class of failure, sent `-32600`
instead. Two codes for one refusal, in one product.

The modern era is already correct: the middleware disengages and the SDK ladder
returns `HEADER_MISMATCH`. This only fires for legacy-era traffic — which is
exactly the traffic that has no other source of truth about why it was refused.

## Reproduction steps

```python
from mcp_hangar import tasks_wire
import inspect, mcp_hangar.fastmcp_server.front_door_routing as fdr

print(tasks_wire.HEADER_MISMATCH)                  # -32020
print(inspect.getsource(fdr._reject))              # was "code": -32600
```

Over the wire: send a legacy-era `POST /mcp` with `Mcp-Method: tools/call` and a
body whose `method` is `prompts/get`. The response is 400 with
`{"error": {"code": -32600, …}}` before the fix.

## Expected behavior

The same refusal carries `HEADER_MISMATCH` on both paths, so a client's error
handling and an intermediary's log parsing do not have to know which of our
modules answered. The SDK already maps this code to HTTP 400
(`mcp/shared/inbound.py:312`), so the status stays where it is.

## Actual behavior

`_reject` (`fastmcp_server/front_door_routing.py`) hard-coded `-32600`
(`InvalidRequest`). Tests pinned that value (`test_front_door_header_routing.py`,
`test_serve_modern_surface.py`).

Same shape as the class named in the "Sixteen Releases" post: the mechanism is
present, two modules away, and this path does not reach it.

## Version

2.13.0. `HEADER_MISMATCH` has been correct in `tasks_wire.py` since the T-2663
work; the front door has carried `-32600` since it was written.

## Severity

p3-low. Nothing is mis-enforced — the refusal happens, fail-closed, with the right
HTTP status. It is a wire-truthfulness defect, in the same family as #890 and
#891, and cheap to fix while #1050 (base64 decode) is open in the same function.

## Fix

Import `HEADER_MISMATCH` from `tasks_wire` (already exported; `task_relay_handlers`
already imports it from there) and use it in `_reject`. No second literal.

## Acceptance

* A legacy-era header/body mismatch returns HTTP 400 with code `-32020`.
* The literal `-32600` no longer appears in `front_door_routing.py`.
* A test asserts front-door and `tasks/*` mismatches carry the **same** code
  (`HEADER_MISMATCH` imported from `tasks_wire` in both test modules).

mapyr:
## Why

Four defects on the SEP-2243 header path, all on the same two functions, so
they land together rather than as four rebases of each other.

**#1049 — the projection is rebuilt per call, and the metric counts it.** The
SDK's modern transport resolves a called tool's `inputSchema` by invoking our
own `tools/list` pre-dispatch, on every `tools/call` that carries arguments.
There is no cache. For the front door that means identity resolution, policy
resolver and `_build_flat_map` run a second time, and `PROJECTED_TOOLS` — the
one number we point at when arguing a governance plane is not free — records
an observation per call instead of per listing. A metric that silently changes
meaning is worse than an absent one, because it gets quoted.

**#1050 — a base64 `Mcp-Name` is refused.** The sentinel is mandatory when the
value is non-ASCII, and `Mcp-Name` for `resources/read` is the URI, so a Polish
resource path is the ordinary case. The handshake-era front door and `tasks/*`
compared raw latin-1 and never decoded. Fail-closed, so nothing leaked; it just
refused correct traffic.

**#1051 — two codes for one refusal.** `tasks_wire.HEADER_MISMATCH` (-32020)
already exists and the task relay uses it; the front door sent -32600.

**#1053 (partial) — the fail-open boundary was silent.** When the nested listing
cannot produce a schema, the SDK skips `Mcp-Param-*` validation and the call
proceeds. Some branches log, none were a metric. An enforcement plane should
know when its enforcement did not run.

While registering the new counter: `PROJECTED_TOOLS` and
`EMPTY_PROJECTION_TOTAL` were defined but never added to
`_register_all_metrics()`, so neither has ever reached `/metrics`. Both are
registered here. The three `APPROVAL_*` counters have the identical defect and
are filed as #1059 rather than smuggled into this diff.

## What changed

- `front_door_routing.py` — routing headers go through `decode_header_value`;
  a malformed sentinel is kept as the raw field so it cannot match a body value
  by accident. `_reject` imports `HEADER_MISMATCH` instead of carrying a second
  literal.
- `task_relay_handlers.py` — same codec before the `taskId` comparison.
- `flat_tool_projection.py` — the identity-scoped projection is memoised on
  `request.state`, keyed on tenant, for one HTTP request; the call handler
  reuses it. `PROJECTED_TOOLS` / `EMPTY_PROJECTION_TOTAL` are observed only for
  a client-visible listing, detected from the cached request body's envelope
  method. Skips are counted by reason.
- `metrics.py` — `PARAM_HEADER_VALIDATION_SKIPPED_TOTAL{reason}`, plus the two
  registrations that were missing.

Not in scope: the fail-open ADR (docs#269, human-authored) and the
`param_validation.required` flag it names, both still open on #1053.
Projection-time `x-mcp-header` validation is #1056.

## How tested

`uv run pytest tests/unit` — 6655 passed, 2 skipped. `ruff format --check`,
`ruff check` and `mypy src` clean (the 5 remaining mypy errors are in
`tracing.py` / `langfuse.py` and predate this branch).

New:

- `test_a_tools_call_does_not_recount_the_projection.py` — a list and a call on
  one request build the map once; a client-visible listing is counted and a
  nested one is not; the pre-dispatch listing resolves the **same principal** as
  the call it validates and fails loudly if it resolves to anonymous (#874
  class). Verified by sabotage: stubbing `bind_caller_identity` to return `None`
  turns both of those red.
- `test_param_header_validation_skips_are_counted.py` — the counter is on the
  exposition; each of the four reasons increments distinctly; a client-visible
  listing is not a skip.
- `test_front_door_header_routing.py` — a sentinel-wrapped `Mcp-Name` matching
  the body routes, one decoding to a different value is refused, whitespace
  wrapping matches the stripped body. Driven over a real ASGI request, not
  through `resolve_route`.
- `test_task_relay_handlers.py` — the same two cases for `tasks/*`.

Fixes #1049
Fixes #1050
Fixes #1051

### DIFF
--- a/changelog.d/1049-sep2243-headers-and-projection-memo.fixed.md
+++ b/changelog.d/1049-sep2243-headers-and-projection-memo.fixed.md
@@ -1,0 +1,7 @@
+**core:** a handshake-era `Mcp-Name` that used the SEP-2243 base64 sentinel was
+refused as a header/body mismatch, and a modern `tools/call` with arguments
+recounted `mcp_hangar_projected_tools` because the SDK's schema lookup re-ran
+`tools/list`. Routing headers now go through `decode_header_value`, a mismatch
+answers `-32020` (`HEADER_MISMATCH`) like `tasks/*`, and the identity-scoped
+projection is memoised for the lifetime of one HTTP request so the metric
+still measures listings a client actually received.
--- a/changelog.d/1053-param-header-validation-skips.fixed.md
+++ b/changelog.d/1053-param-header-validation-skips.fixed.md
@@ -1,0 +1,5 @@
+**core:** a `tools/call` whose `Mcp-Param-*` headers were never checked left no
+metric. Hangar now increments `mcp_hangar_param_header_validation_skipped_total`
+when the nested listing fails, omits the tool, or advertises an invalid
+`x-mcp-header`, and when a handshake-era call still carries `Mcp-Param-*`.
+The fail-open boundary itself is unchanged.
--- a/src/mcp_hangar/fastmcp_server/flat_tool_projection.py
+++ b/src/mcp_hangar/fastmcp_server/flat_tool_projection.py
@@ -51,7 +51,7 @@ import hashlib
 import json
 import logging
 import uuid
-from typing import Any, cast
+from typing import Any
 
 from mcp.shared.inbound import MCP_PARAM_HEADER_PREFIX, find_invalid_x_mcp_header
 
@@ -362,11 +362,12 @@ EMPTY_NO_IDENTITY = "no_identity"
 EMPTY_NOTHING_DISCOVERED = "nothing_discovered"
 EMPTY_FILTERED = "filtered"
 
-# Per-HTTP-request memo of the identity-scoped projection. The SDK's modern
+# Per-HTTP-request memo of the identity-scoped flat map. The SDK's modern
 # transport re-invokes tools/list pre-dispatch to resolve Mcp-Param schemas
-# (#1049); without this, that listing rebuilds the map and recounts the metric.
+# (#1049); without this the call rebuilds the map that listing just built.
 _MEMO_ATTR = "hangar_projection"
-_ENVELOPE_ATTR = "hangar_envelope_method"
+
+_MCP_PARAM_PREFIX_LOWER = MCP_PARAM_HEADER_PREFIX.lower()
 
 
 def _http_request(mcp_ctx: Any) -> Any:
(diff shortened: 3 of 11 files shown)


mapyr:
## Ponytail-review — possible cuts

Header decode, `-32020`, and the skip counter itself stay. The cut is the helper pile around memo + envelope in `flat_tool_projection.py`.

`flat_tool_projection.py:L377-426:` **shrink:** `_envelope_method` and `_envelope_payload` both `json.loads` `request._body`; method is then cached on `_ENVELOPE_ATTR` with `try/except setattr`. One `_envelope()` returning the dict, no cache — Starlette already buffered the body.

`flat_tool_projection.py:L409-411:` **yagni:** `_client_visible_listing`. `(_envelope(ctx) or {}).get("method") != "tools/call"` at the one call site.

`flat_tool_projection.py:L429-453:` **yagni:** `_called_tool_name`, `_header_keys`, `_input_schema_of`. One caller each. Inline; prefix scan is `any(str(k).lower().startswith(...) for k in headers or ())`; schema is `tool.inputSchema`.

`flat_tool_projection.py:L469-472:` **yagni:** `_observe_listing_failed_skip`. The two-line `if` belongs in the `except`.

`flat_tool_projection.py:L486-559:` **yagni:** five functions to memo a dict for one list-then-call POST. SDK lists once — drop the list-path cache hit, the stored `ListToolsResult`, and `load_management`. `setattr(state, _MEMO_ATTR, (tenant_id, flat_map))` in the list handler; read it in the call handler; call `_management_tools` directly. Drop `except Exception: pass` around `setattr` — Starlette `request.state` always takes it.

`front_door_routing.py:L80-85:` **shrink:** `raw = value.decode("latin-1").strip()` then `return (decode_header_value(raw) or raw).strip() or None`.

net: **-80 lines possible.**

mapyr:
Taken, with two corrections and one that the ratchet vetoed. Net **-84 lines** in `b1bcf98`.

**Applied as written**

- `_envelope_method` + `_envelope_payload` → one `_envelope()` returning the parsed dict. The `_ENVELOPE_ATTR` cache and its `try/except setattr` are gone: `json.loads` of one small buffered body a few times per request is not worth a second state bag.
- `_client_visible_listing`, `_called_tool_name`, `_observe_listing_failed_skip` inlined at their single call sites.
- `_header_keys` had two callers, not one, so the prefix scan became `_carries_param_header(ctx)` shared by both rather than two inline copies.
- Memo now stores `(tenant_id, flat_map)`. You are right that the list-path cache hit was unreachable — one nested listing per POST, and the front door returns an unpaged list so the SDK stops after one page. `_cached_projection` / `_store_projection` / `_flat_map_for_request` → `_memoised_flat_map` / `_memoise_flat_map`, no `except Exception: pass` around the setattr.

**Corrected**

`_input_schema_of` → `match.input_schema`, not `match.inputSchema`. On the SDK v2 `Tool`, `input_schema` is the field and `inputSchema` is only the serialization alias, so `tool.inputSchema` raises `AttributeError` — pydantic does not expose aliases as attributes. The helper deserved to die; the fallback half was simply never the one that fired here.

`front_door_routing._decoded_routing_header` shrank to three lines, but not to `(decode_header_value(raw) or raw).strip()`. `decode_header_value` returns `None` for a malformed sentinel **and** an empty string for one that decodes to nothing; `or raw` conflates the two and would send the literal `=?base64?B??=` downstream as a present header rather than treating it as absent. Kept as `(raw if decoded is None else decoded).strip() or None`.

**Vetoed by the ratchet**

Inlining the listing into the `_flat_list_tools` closure so `_management_tools` is called directly puts `register_flat_tool_handlers` at C901 18 > 15. `_list_projected_tools(mcp_ctx, load_management)` stays module-level — that indirection is what the complexity gate is buying.

Behaviour unchanged: same four skip reasons, same counted listings, same map built once per POST. `tests/unit` 6655 passed, and the identity sabotage check (`bind_caller_identity` → `return None`) still reddens both projection tests.

Issue #983, opened 2026-08-16
UPGRADE.md accumulates `## Next` sections that no release ever promotes


mapyr:
## What

`UPGRADE.md` in this repo has **36 sections, 8 of them headed `## Next`**. Its newest *versioned* section is **2.6.0**. We have released 2.7.0, 2.8.0 and 2.9.0 since.

Nothing in `.github/workflows/` or `scripts/` reads `UPGRADE.md`, so nothing notices.

## Why it matters

The changelog entries for 2.8.0 and 2.9.0 say *"see `UPGRADE.md`"* — eight times across the removals of `MCPServerFactory`, `LogAuditStore`, `CallbackAlertSink` and `detect_runtime_availability`. A reader who follows that pointer lands on a section headed **"Next"**, with no way to tell which release it describes or whether it has shipped.

Worse, the sections go stale once merged. The `builder()` section says *"`MCPServerFactory`, `HangarFunctions` and `ServerConfig` are unchanged by this release"* — true when it was drafted for #963, false by the time #965 landed in the same release.

## The real shape

There are **two hand-maintained upgrade documents**:

| | newest section | who reads it |
|---|---|---|
| `UPGRADE.md` (here) | 2.6.0 + 8 × `Next` | authors, at PR time |
| `docs/upgrade.md` ([mcp-hangar/docs](https://github.com/mcp-hangar/docs)) | 2.7.0 → 2.9.0 after mcp-hangar/docs#227 | readers, published to the site |

They have diverged. The 2.7.0 session-id section exists in both, promoted in one and still headed `Next` in the other — so the copy happened once, by hand, and the source was never cleaned up.

mcp-hangar/docs#227 adds a gate on the published side: the released minor must have a section. That closes the reader-facing hole. It does not stop this file drifting.

## Options

1. **Promote at release.** `release-please` already rewrites `CHANGELOG.md` and `pyproject.toml`; a step that renames `## Next — X` to `## X.Y.Z — X` is small. Does not fix the divergence.
2. **Delete this file, keep the docs one.** Authors write the section in the docs repo at PR time. One document, already gated, already published. Costs a cross-repo PR for every change that removes API.
3. **Delete the docs copy, publish this one.** The docs site already syncs content from other repos. Authors keep writing where the code is; readers get the same file. Needs the promotion step from option 1 either way.

Option 3 is the one I would take — it keeps the note next to the change that motivates it, which is why the drafts get written at all — but it is a workflow decision, not a code one.

[...]

PR #1067, opened 2026-08-24
docs(core): upgrade notes and a release summary for 2.14.0

mapyr:
## Why

The release PR for 2.14.0 (#1062) is open and green, and `UPGRADE.md` stops at
**2.12.0**. This release changes three things an operator sees without touching
their configuration, and none of them is written down anywhere they would look.

I own that gap: my four PRs in this batch added changelog fragments and no
upgrade notes.

## What changed

`UPGRADE.md` gains an "Upgrade to 2.14.0" section:

1. **A front-door tool can disappear** (#1056). If an upstream ships a malformed
   `x-mcp-header` annotation, a tool listed yesterday is gone today with no
   config change. The note says where the reason is — the
   `tool_withheld_invalid_x_mcp_header` log line and
   `mcp_hangar_projection_withdrawals_total{reason="invalid_x_mcp_header"}` —
   what makes an annotation valid, and why Hangar withholds rather than strips
   (stripping moves the digest and breaks every pin).
2. **A legacy-era header/body mismatch answers `-32020`, not `-32600`** (#1051).
   Nothing that was refused is accepted now, but a client branch or log query
   keyed on the old code needs updating. Modern clients were already on `-32020`.
3. **Eight metrics reach `/metrics` for the first time** (#1059, #1049), listed
   by name. A panel that has been quietly returning no data starts returning
   rows; the approval queries in the observability guide — written against
   counters dead since 2.10.0 — work now.

Plus a short section each for the two new optional controls, including the part
worth knowing this week: the `Mcp-Param-*` selector is **REST-only** until the
operator ships the matching CRD field
([operator#160](https://github.com/mcp-hangar/mcp-hangar-operator/issues/160)),
so a GitOps rollout cannot express it yet.

`changelog.d/_summary.md` gives the release its intro paragraph. This one has a
single theme, and two sentences carry it better than eleven bullets do.

## How tested

`make changelog-check` — 6 fragments valid. `make changelog-preview
VERSION=2.14.0` renders the summary above the sections as intended.

Every `mcp_hangar_*` name in the new section was checked against the live
registry rather than typed from memory — all eight resolve, which is the check
that matters given the bug being described is that they were *absent*.

## Merge order

This wants to land **before** #1062 is merged: release-please will regenerate
that PR to include the summary, and the upgrade guide should ship with the tag,
not after it. `sync-upgrade-guide.yml` then carries the section into the docs
repo.

Pure docs — no changelog fragment needed per `changelog.d/README.md`.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,88 @@
 # Upgrading MCP Hangar
 
+## Upgrade to 2.14.0
+
+### A front-door tool can disappear, with no config change (#1056)
+
+SEP-2243 makes it a client-side **MUST**: a client drops any tool whose
+`x-mcp-header` annotations are invalid. Hangar forwarded the upstream definition
+verbatim, so it advertised such a tool and counted it as surface delivered while
+every conforming client silently dropped it.
+
+Since this release the tool is withheld at the projection instead: absent from
+`tools/list`, and `-32601` on the call. **If an upstream ships a malformed
+annotation, a tool that was listed yesterday is gone today**, and nothing in your
+configuration changed. The reason is in the log
+(`tool_withheld_invalid_x_mcp_header`, naming the tool and what was wrong) and on
+`mcp_hangar_projection_withdrawals_total{reason="invalid_x_mcp_header"}`. Scrape
+that counter after the upgrade if you want to know before your users do.
+
+An annotation is valid only when it sits on a property reachable through a pure
+`properties` chain, names an RFC 9110 token, is on an `integer`/`string`/
+`boolean` property, and is unique across the schema. The fix belongs upstream;
+Hangar does not edit a projected schema, because that would move its digest and
+break every pin.
+
+### A legacy-era header/body mismatch now answers `-32020` (#1051)
+
+The handshake-era front door refused a mismatch with `-32600` (`InvalidRequest`)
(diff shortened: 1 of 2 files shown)


mapyr:
Corrected in `26eac9c` — the first push was wrong and CI was right.

`test_the_upgrade_notes_get_the_version_they_shipped_in` refuses an
`## Upgrade to <version>` heading past the released one, and 2.14.0 does not
exist yet. A section stamped with an unreleased version reads as shipped, which
is exactly the failure #983 recorded: eight `Next` sections accumulated across
2.7.0–2.9.0 because nothing gave them a version, and the fix was to promote them
**at release time**, not to stamp them by hand at PR time.

So the notes are now five `## Next — <headline>` drafts.
`assemble_release_changelog.sh` folds them into `### <headline>` under
`## Upgrade to 2.14.0` on the release branch. A local dry run of
`promote_upgrade_notes.promote` against a copy produces exactly that, in order,
with no `Next` left behind.

Worth naming the mistake rather than just the fix: I pushed this without running
the suite because it was "just docs", in a repo where the tests read the docs.
`make changelog-check` passing is not the same as `pytest` passing.

The `decision-coverage` failure was a knock-on — that job runs the same suite
under `--cov`, so the one red test failed it too. Both should be green now;
`tests/unit` is 6717 passed locally.

Issue #1101, opened 2026-08-28
epic: fuzz the policy evaluator


mapyr:
Scorecard reports `Fuzzing: 0`. The check is binary (Medium, weight 5), so
0 -> 10 is **+0.51** on the total -- second only to Token-Permissions.

That is not why this is being written. Surveying the surface for a harness
turned up a real defect in twenty minutes by hand (#1102), which is the argument
that an automated one will find more.

## Children

- [x] #1102 `RecursionError` escapes `evaluate()` instead of a verdict
- [x] #1103 an atheris harness over the policy evaluator
- [ ] #1104 ClusterFuzzLite on pull requests

Plus #1106, found while writing #1103's precedence invariant: `merge()` dropped
a `deny_list`, so a tool denied at the server scope came back allowed once a
group scope repeated the allow list. A policy bypass, closed by #1107.

## Order

#1102 is a standalone fix and goes first; it is worth nothing on the Scorecard and
is the most important of the three. #1103 is where the points actually come from
(`import atheris` in the tree). #1104 turns a one-off run into a standing check.

## What we are not doing

Scorecard detects Python fuzzing **only** by grepping for `import atheris` in
any `.py` file (`checks/raw/fuzzing.go`, `clients.Python`). Hypothesis and
property-based testing are not recognised for Python -- Erlang, Haskell, Elixir,
Gleam, JS, TS, C# and F# have detectors, Python does not.

So an empty file with one import scores a full 10 and +0.51. **We are not doing
that.** It is the same move as a bot approving its own pull requests: it costs
half a day less and the entire credibility of a product whose one promise is
that the verdict is deterministic and attributable.

If the harness does not get built, this epic closes as wontfix and `Fuzzing: 0`
stands. Zero is honest.

mapyr:
All four children closed.

| # | | |
|---|---|---|
| #1102 | `RecursionError` escaped `evaluate()` instead of a verdict | #1105 |
| #1106 | `merge()` dropped a `deny_list`, so a denied tool came back allowed | #1107 |
| #1103 | atheris harness over the policy evaluator | #1108 |
| #1104 | ClusterFuzzLite on pull requests | #1111 |

**The epic predicted one bug and produced two**, both before a fuzzer ever ran
-- writing the invariants down was what found them. #1106 was not in the plan
at all: a policy **bypass**, reachable from an ordinary two-level
server/group configuration, and exactly the class the plan said the precedence
[...]

PR #1105, opened 2026-08-28
fix(core): make the policy evaluator total

### LINKED ISSUE #1102: [BUG] RecursionError escapes evaluate() instead of a verdict
Part of #1101.

## Reproduction (confirmed on `main`, not only on the 2.14.1 wheel)

```python
from mcp_hangar.domain.policies.egress_l7 import L7Policy, evaluate

pol = L7Policy.from_dict({
    "mode": "enforce",
    "tools": {"allow": ["some_tool"], "defaultAction": "deny"},
    "arguments": {"secretPatterns": ["aws-keys"], "maxPayloadBytes": 100000},
})

def nest(n):
    d = {}; cur = d
    for _ in range(n):
        cur["a"] = {}; cur = cur["a"]
    return d

evaluate("some_tool", nest(100), pol)   # Decision(action=ALLOW)
evaluate("some_tool", nest(992), pol)   # RecursionError
```

**The threshold is lower than the report estimated.** A binary search puts the
first raising depth at **992**, which is **6 946 bytes** of JSON -- not the
20 000 levels / ~40 KB in the brief. Any `maxPayloadBytes` an operator would
plausibly set is far above it, so the size guard is not merely bypassed in
theory: 7 KB is an ordinary tool call.

**Audit mode was confirmed by running it, not derived.** The brief flagged this
as needing confirmation:

```python
audit_pol = L7Policy.from_dict({"mode": "audit", ...})
evaluate("some_tool", nest(992), audit_pol)   # RecursionError
```

ADR-013 says audit observes and lets the call through -- the safe adoption path.
It does not: the exception aborts the call before dispatch whatever the mode, so
an operator switching a policy on in audit gets a hard failure instead of an
observation.

## Mechanism

`_serialize_arguments` (`egress_l7.py:367`):

```python
try:
    return json.dumps(arguments, sort_keys=True, ..., default=str)
except (TypeError, ValueError):
    return None
```

The docstring promises `None` when the payload cannot be serialized "so the
caller can fail closed rather than crash". `RecursionError` is neither
`TypeError` nor `ValueError`, so the promise does not hold. It propagates
through `scan_arguments` and `evaluate`, and `mcp_server.py:1282` calls
`evaluate(...)` with no `try` around it.

Two consequences worth naming:

- **`maxPayloadBytes` cannot protect against this** -- the size check runs
  *after* serialization, so the guard sits behind the thing that breaks.
- **In enforce, fail-closed happens by accident.** It comes out of the order of
  operations, not a decision: no `EgressPolicyDeniedError`, no
  `EgressPolicyViolationObserved`, no audit entry. The call dies as an
  unattributed error. For a plane whose thesis is that every tool call ends in
  a verdict, this is a call that ends in a stack trace.

`default=str` is a second door onto the same problem: it invokes arbitrary
`__str__`, which is somebody else's code and may raise anything at all.

## Change

- `_serialize_arguments` catches broadly rather than by name, and logs what it
  caught. `RecursionError` is a `RuntimeError`, so `except Exception` covers it
  and the `default=str` case in one move.
- `evaluate()` becomes total: no exception from argument inspection escapes it.
  In enforce that is a DENY with a reason; in audit the caller then observes and
  proceeds, which is what ADR-013 promised.

## Acceptance

- [ ] `evaluate()` returns a `Decision` at nesting 10, 1 000 and 100 000
- [ ] enforce: DENY with a readable reason, `EgressPolicyDeniedError` raised
- [ ] audit: the call proceeds, `EgressPolicyViolationObserved` emitted
- [ ] the reproduction above lands as a regression test
- [ ] a changelog fragment describing the mechanism, not the PR title

mapyr:
## Why

A tool call whose arguments cannot be serialized ends in a stack trace instead
of a verdict.

```python
evaluate("some_tool", nested(992), policy)   # RecursionError
```

`_serialize_arguments` caught `(TypeError, ValueError)` and promised `None` "so
the caller can fail closed rather than crash". `RecursionError` is neither, so
it propagates through `scan_arguments`, out of `evaluate()`, and out of
`McpServer._enforce_l7_policy`, which calls it with no `try`.

Closes #1102. Refs #1101.

## Two corrections to the report

- **The threshold is 992 levels, not 20 000.** A binary search on `main` puts
  the first raising depth at 992, which is **6 946 bytes** of JSON. The brief
  estimated ~40 KB. 7 KB is an ordinary tool call, and `maxPayloadBytes` cannot
  help at any setting: the size check reads the string the serializer returns,
  so the guard sits behind the thing that breaks.
- **Audit mode was confirmed, not derived.** The brief marked it as needing an
  end-to-end check. Running `evaluate` with `mode: audit` raises the same way,
  and the new test drives it through `invoke_tool`: before this fix, a policy
  switched on in the mode ADR-013 documents as the *safe adoption path* was a
  hard block on every call with deep arguments.

In Enforce the call was refused, but by accident of ordering rather than by
decision -- no `EgressPolicyDeniedError`, no `EgressPolicyViolationObserved`,
no audit entry. The call died unattributed, in a plane whose thesis is that
every call ends in an attributable verdict.

## What

- `_serialize_arguments` catches by class rather than by name. `RecursionError`
  is a `RuntimeError`, so one `except Exception` also covers `default=str`,
  which invokes an arbitrary `__str__` -- the second door onto the same
  problem. The reason is logged; the caller's verdict stays the deliberately
  unspecific "could not be serialized", because what belongs in an audit record
  is that the call was refused *uninspected*.
- `evaluate()` is total by contract: inspection that fails for any reason
  becomes a DENY with a reason. Enforce then raises as it always meant to, and
  Audit records and lets the call through.

A depth pre-check was considered and left out: catching is what makes the
function total, and a second traversal to produce a slightly better reason
string is not worth its own recursion bug.

## How tested

- `tests/unit/test_the_policy_evaluator_always_returns_a_verdict.py` (new, 11
  tests): a verdict at depth 10 / 1 000 / 100 000 in both modes, shallow calls
  still ALLOW (the fix must not turn ordinary calls into denials), an object
  whose `__str__` raises, Enforce raising `EgressPolicyDeniedError` with the
  reason, and Audit emitting `EgressPolicyViolationObserved` while the call
  proceeds.
- Probe: with only `egress_l7.py` reverted, 8 of the 11 fail.
- `pytest tests/unit` -- 6750 passed, 2 skipped. `ruff` clean, `mypy src` has
  the same 5 pre-existing errors and no new ones.

## Risk and rollback

Arguments that previously crashed now DENY in Enforce. That is the intended
change and matches the documented contract; a caller relying on the crash was
relying on an unattributed failure. Revert the single commit.

## Agent metadata

- [x] LOC declared upfront: `~30` excluding tests

### DIFF
--- a/changelog.d/1102-policy-evaluate-is-total.fixed.md
+++ b/changelog.d/1102-policy-evaluate-is-total.fixed.md
@@ -1,0 +1,22 @@
+**core:** a tool call whose arguments could not be serialized ended in a stack
+trace instead of a verdict. `_serialize_arguments` caught `TypeError` and
+`ValueError` and promised `None` "so the caller can fail closed rather than
+crash", but JSON nesting the encoder cannot walk raises `RecursionError`, which
+is neither -- about 992 levels, roughly 7 KB of payload, well under any
+`maxPayloadBytes` an operator would set. `maxPayloadBytes` could not have
+helped either way: the size check reads the string the serializer returns, so
+the guard sat behind the thing that broke. The exception propagated out of
+`evaluate()` and out of `McpServer._enforce_l7_policy`, which called it with no
+`try`.
+
+In Enforce the call was refused, but by accident of ordering rather than by
+decision: no `EgressPolicyDeniedError`, no observation, no audit entry -- the
+call died unattributed. In **Audit** the exception aborted the call before
+dispatch, so a policy switched on in the mode ADR-013 documents as the safe
+adoption path was a hard block instead of an observation.
+
+`evaluate()` is now total: argument inspection that fails for any reason
+produces a DENY carrying a reason, which Enforce raises on and Audit records
+and lets through. The failure is logged with the underlying error. The same fix
+covers `default=str`, which invokes an arbitrary `__str__` and can raise
+anything.
--- a/changelog.d/1102-policy-evaluate-is-total.fixed.md
+++ b/changelog.d/1102-policy-evaluate-is-total.fixed.md
@@ -2,8 +2,10 @@
 trace instead of a verdict. `_serialize_arguments` caught `TypeError` and
 `ValueError` and promised `None` "so the caller can fail closed rather than
 crash", but JSON nesting the encoder cannot walk raises `RecursionError`, which
-is neither -- about 992 levels, roughly 7 KB of payload, well under any
-`maxPayloadBytes` an operator would set. `maxPayloadBytes` could not have
+is neither. The depth that triggers it is an interpreter detail rather than a
+constant: 992 levels on CPython 3.11 -- this project's baseline, so roughly
+7 KB of payload, far under any `maxPayloadBytes` an operator would set -- and
+9 997 on 3.12, 34 710 on 3.14. `maxPayloadBytes` could not have
 helped either way: the size check reads the string the serializer returns, so
 the guard sat behind the thing that broke. The exception propagated out of
 `evaluate()` and out of `McpServer._enforce_l7_policy`, which called it with no
--- a/src/mcp_hangar/domain/policies/egress_l7.py
+++ b/src/mcp_hangar/domain/policies/egress_l7.py
@@ -28,11 +28,14 @@ from dataclasses import dataclass, field
 from enum import StrEnum
 from fnmatch import fnmatchcase
 import json
+import logging
 import re
 from typing import Any
 
 from ..._sdk_compat import is_modern_protocol_version
 from ...redactor import OutputRedactor
+
+logger = logging.getLogger(__name__)
 
 
 class ToolAction(StrEnum):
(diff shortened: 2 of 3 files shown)


mapyr:
CI was red on 3.12 and my test was wrong, not the fix.

The tests hard-coded 1 000 levels of nesting. That is the threshold on **3.11
only**. Measured on each interpreter this project tests:

| CPython | first depth `json.dumps` refuses |
|---------|----------------------------------|
| 3.11 | 992 |
| 3.12 | 9 997 |
| 3.14 | 34 710 |

So on 3.12+ the test fed 1 000 levels, the encoder serialized them happily, the
policy allowed the call, and the assertions failed -- the test passed on 3.11 by
reproducing the bug and failed elsewhere by not reproducing it. The two
downstream failures (`McpServerStartError: Executable 'echo' is not in the
allowed command list`, and `assert 0 == 1` on the observation) were the same
cause: the call was never blocked, so it ran on to `ensure_ready`.

Fixed by asking the local encoder where its own edge is -- doubling from 512
until `json.dumps` raises, skipping if it never does. Verified on 3.11, 3.12,
3.13 and 3.14 locally: 11 passed on each.

This also corrects the issue text and the changelog fragment, both of which
quoted "992 levels, roughly 7 KB" as if it were a constant. It is 3.11's
number. 3.11 is this project's baseline (`.python-version`) so it remains the
worst case and the honest headline, but a deployment on 3.14 needs ~35 000
levels to hit it.

Issue #1106, opened 2026-08-28
[BUG] merge() drops a deny_list, so a denied tool comes back allowed


mapyr:
Found while writing the precedence fuzz target for #1103, on the first handful
of inputs. This is the class the epic predicted that target could find: a
**policy bypass**, not a crash.

## Reproduction

```python
from mcp_hangar.domain.value_objects.tool_access_policy import ToolAccessPolicy

broader  = ToolAccessPolicy(allow_list=["*"], deny_list=["drop_*"])  # server scope
narrower = ToolAccessPolicy(allow_list=["*"])                        # group scope

broader.is_tool_allowed("drop_db")                       # False
ToolAccessPolicy.merge(broader, narrower).is_tool_allowed("drop_db")   # True
```

A tool denied at the server scope comes back allowed after merging with a group
scope that does nothing but repeat the allow list. The narrower scope did not
lift the denial -- it cannot, by design -- it just made it disappear.

## It breaks the invariant `merge` documents about itself

`merge`'s own docstring states:

```
merged.filter_tools(tools) == narrower.filter_tools(broader.filter_tools(tools))
for ALL possible tool lists
```

It does not hold:

```python
tools = ["read_file", "drop_db"]
merged.filter_tools(tools)                          # ['read_file', 'drop_db']
narrower.filter_tools(broader.filter_tools(tools))  # ['read_file']
```

## Mechanism

`merge` dispatches on which lists are populated, and each branch hand-rolls a
check out of the lists it names -- dropping the ones it does not.

* **Case 3** (`tool_access_policy.py:219`, both sides have an `allow_list`)
  builds `allow(broader) and allow(narrower)` and never consults either
  `deny_list`. This is the reproduction above.
* **Case 4** (broader `allow` + narrower `deny`) reads `broader.allow_list` and
  `narrower.deny_list`, dropping `broader.deny_list` if it has one.
* **Case 5** (broader `deny` + narrower `allow`) is the mirror image: it drops
  `narrower.deny_list`.

So all three branches lose a deny list; only the shape of the input decides
which one. Case 6 (both deny) and case 7 (approval) are correct.

The deny > approval > allow > default ladder is already implemented, correctly,
in `is_tool_allowed`. These branches reimplement a piece of it from raw lists,
which is why they can disagree with it.

## Change

Compose the two policies' own decisions instead of their lists: the merged
check is `broader.is_tool_allowed(name) and narrower.is_tool_allowed(name)`.
[...]

Issue #1104, opened 2026-08-28
ci: ClusterFuzzLite on pull requests


mapyr:
Part of #1101. Follows #1103.

A single fuzzing run is an experiment, not a control. ClusterFuzzLite runs a
short fuzz over the diff on every pull request and carries the corpus between
runs.

## Change

- `.github/workflows/cfl-pr.yml` -- batch mode on PR, a budget of a few minutes
- `.clusterfuzzlite/` with the build configuration
- corpus cached between runs
- a crash becomes an Actions artifact and fails the job

## Acceptance

- [ ] a PR with a deliberately planted crash is caught
- [ ] the job stays inside the agreed CI time budget
- [ ] after a re-scan: `Fuzzing = 10`, total ~7.7

mapyr:
Done in #1111, with the acceptance criterion demonstrated rather than assumed.

**A PR with a deliberately planted crash is caught.** A `RuntimeError` planted
in `evaluate_tool` was found by `mode: code-change` in under three minutes,
reproduced (`INFO - Crash is reproducible.`), written out as a testcase and
failed the job. The plain atheris job from #1103 found the same crash in 20
seconds -- two independent layers. Both commits stay in #1111's history as the
evidence.

**The job stays inside the budget.** 180 seconds of fuzzing, ~20 minutes
end-to-end including the OSS-Fuzz build, on a path filter that only fires when
`fuzz/`, the policy modules or the CFL config change.

**`Fuzzing = 10`** since #1108 -- Scorecard reads `import atheris` from the
tree, so it was already there before this landed. The total is 8.1.

One correction to the plan: the issue and my first draft both said corpus
persistence needed a separate storage repo. It does not. CFL's default
filestore is GitHub Actions artifacts, which the first run showed by looking
for `cifuzz-corpus-fuzz_policy_evaluate` and not finding it. What was missing
was the job that produces them, so `cfl-batch.yml` ships alongside: 30 minutes
nightly in batch mode. A PR-only setup would have been a fuzzer with amnesia,
and I would have shipped exactly that on the strength of a guess.

PR #1108, opened 2026-08-28
test(core): fuzz the policy evaluator

### LINKED ISSUE #1103: test: an atheris harness over the policy evaluator
Part of #1101. Follows #1102.

## Goal

Fuzzing looks for an input where the decision function **does not produce a
decision**. Not "the answer should be X" assertions -- invariants whose breach
is by definition a security bug.

## Targets

| Module | Input | Invariant |
|---|---|---|
| `domain/policies/egress_l7.py::evaluate` | `tool_name`, `arguments: Any`, policy, headers | Always returns a `Decision`. Never propagates. |
| `domain/policies/egress_l7.py::L7Policy.from_dict` | `dict[str, Any]` from the operator / CRD | Either a valid object or a declared validation error. Never `KeyError` / `AttributeError` / `RecursionError`. |
| `domain/policies/dsl.py::parse_policy` | `data: Any` | As above. |
| `domain/value_objects/tool_access_policy.py::ToolAccessPolicy` | tool name + pattern sets | **Precedence**: if the name matches `deny_list`, `is_tool_allowed` is `False` regardless of the other lists -- including after `merge()` and in `_CompositePolicy.merge_with()`. |
| `domain/services/digest_validator.py` | a digest from an upstream | An invalid digest is rejected, never raised over. |
| `application/mutators/response_truncator.py` | an upstream response | Truncation neither loses nor corrupts the response boundary. |

The fourth is the interesting one. `_matches_any_pattern` rests on `fnmatch`,
and the deny > approval > allow > default precedence is computed across
`merge` and the composite class. A fuzzer generating (tool name, pattern sets)
pairs and asserting that deny always wins is the only target on this list that
can find a **policy bypass** rather than a crash.

## Regex cost

`scan_arguments` runs agent-controlled payload through `SECRET_PATTERN_GROUPS`
(`aws-keys`, `jwt`, `pem-blocks`, `github-tokens`, `stripe-keys`,
`slack-tokens`, `google-api-keys`, `bearer-tokens`, `npm-tokens`,
`pypi-tokens`). Ten regex groups over external input is a candidate for
catastrophic backtracking, so the harness needs a time budget: a single
`evaluate()` under a fixed wall-clock threshold, over it is a failure.

## Change

- `fuzz/fuzz_policy_evaluate.py`, `fuzz/fuzz_policy_parse.py`,
  `fuzz/fuzz_access_precedence.py`
- `atheris.Setup` + `FuzzedDataProvider`, with argument structure built from the
  bytes (not flat strings -- otherwise nesting is never reached, which is
  exactly the shape #1102 turned out to be)
- `pyproject.toml`: a `[fuzz]` extra carrying `atheris`
- a seed corpus in `fuzz/corpus/` taken from the real policies under `examples/`
- `fuzz/README.md`: running it locally, reproducing a crash from an artifact

## Acceptance

- [ ] `python fuzz/fuzz_policy_evaluate.py -runs=1000000` clean, or its findings
      filed as issues
- [ ] the #1102 reproduction sits in the corpus as a regression
- [ ] `import atheris` present in a real harness, not a decoy (see #1101)

mapyr:
## Why

Fuzzing looks for an input where the decision function does not produce a
decision -- or produces one answer through one path and a different one through
another. Not "the verdict should be X"; the verdict is the policy's business.

Closes #1103. Refs #1101.

## It has already found two bugs, before the fuzzer ran

Both from writing the invariants down and feeding them a handful of
hand-written cases:

- **#1102** (merged) -- `RecursionError` escaped `evaluate()` instead of a
  verdict. In Enforce the call died unattributed; in Audit it was blocked
  outright, the opposite of what ADR-013 promises Audit does.
- **#1106** (merged) -- `merge()` dropped a `deny_list`, so a tool denied at
  the server scope came back allowed once a group scope repeated the allow
  list. A policy **bypass**, which is the class the epic predicted the
  precedence target could find.

## What

| File | Invariant |
|---|---|
| `fuzz/fuzz_policy_evaluate.py` | `evaluate()` returns a `Decision`, in bounded time |
| `fuzz/fuzz_policy_parse.py` | `L7Policy.from_dict` / `dsl.parse_policy` reject with `ValueError` or accept -- never `KeyError`, `AttributeError`, `RecursionError` |
| `fuzz/fuzz_access_precedence.py` | deny wins, and keeps winning after `merge()` in either order |

Plus `fuzz/corpus/parse/` (7 readable seeds), `fuzz/README.md`, a `[fuzz]`
extra, a bounded CI job, and `tests/unit/test_the_fuzz_invariants_hold.py`
(24 tests) replaying the invariants on every PR.

## Two structural decisions worth reviewing

**The invariants do not import atheris.** `atheris` publishes manylinux x86_64
wheels for CPython **3.12-3.14 only** -- so it installs on neither this
project's 3.11 baseline (`.python-version`) nor any macOS machine. Had the
invariants lived inside the harnesses they would have been checkable in exactly
one environment. They live in `fuzz/invariants.py`; the harnesses are thin
wrappers; the unit suite replays them everywhere in under a second. The extra is
markered (`sys_platform == 'linux' and python_version >= '3.12' and
platform_machine == 'x86_64'`) so `uv sync --all-extras` keeps working here --
verified locally.

**A finding becomes a named test, not a corpus blob.** `fuzz_policy_evaluate`
and `fuzz_access_precedence` consume raw bytes through `FuzzedDataProvider`, so
a hand-written seed for them is unreadable. libFuzzer builds that corpus itself
(and #1104 will persist it). What a human should be able to read is the
regression, so #1102's deep nesting and #1106's bypass are cases in the replay
test.

The `evaluate` harness deliberately builds *structure* from the input bytes,
including an explicit deep-chain branch. A flat-string fuzzer would never have
reached #1102, which was entirely about nesting.

## How tested

- `pytest tests/unit tests/ci` -- 6858 passed, 2 skipped. `ruff` clean,
  `actionlint` clean.
- The 24 invariant tests pass now that #1102 and #1106 are in `main`; on the
  pre-fix tree they failed, which is how both bugs surfaced.
- **The atheris harnesses themselves have not run on my machine and cannot**
  (macOS/3.11, no wheel). `.github/workflows/fuzz.yml` runs each target for
  20 000 iterations on ubuntu/3.12 -- this PR's own run is the first execution,
  and the honest place to find out whether the wrappers work. Shipping three
  harnesses nobody had executed would have been the "advertised but never
  wired" shape this repository keeps finding, so the job is part of the PR
  rather than deferred to #1104.
- `-runs=1000000` from the issue's acceptance is not what CI does: the job is a
  smoke test of the harness, and a real campaign needs a corpus that persists
  between runs, which is #1104.

## Risk and rollback

Adds files and one workflow; changes no production code. The `[fuzz]` extra is
markered so it cannot break an install. Revert the single commit.

## Agent metadata

- [x] LOC declared upfront: `~120` excluding tests and corpus


mapyr:
Both failures were mine, and neither was in the code under test.

**The fuzz job worked on its first run.** `fuzz_policy_evaluate` and
`fuzz_policy_parse` did 20 000 iterations clean. `fuzz_access_precedence` found
an input on its **second** iteration, wrote `crash-00b28ff0...`, and the
artifact upload picked it up -- so the crash-to-artifact path is proven rather
than assumed. What it found:

```
ValueError: Invalid allow_list pattern: ''
  fuzz/invariants.py:100 in check_access_precedence
  tool_access_policy.py:71 in __post_init__
```

`ToolAccessPolicy` rejects an empty pattern by design. That is a declared
outcome, exactly like a policy `L7Policy.from_dict` refuses -- and
`check_evaluate` already skipped those, while `check_access_precedence` did
not. So the harness reported its own bug as a product finding. Fixed on both
ends: the invariant skips a policy the constructor refuses, and `_word` no
longer generates empty patterns, which was spending the budget on a declared
error instead of on the precedence question.

Worth keeping: this is what the target is *for* -- it took two iterations to
find a case a human writing seeds by hand had not.

**The changelog gate failed a `test(core):` PR** for touching `pyproject.toml`
(the `[fuzz]` extra). Same class as #1078: `changelog.d/README.md` lists `ci`,
`style`, `test` and pure `docs` as owing no fragment alongside `chore(deps)`,
and the gate implemented only the last one. Each of those kinds can touch a
triggering path -- a test PR adding an extra, a CI change beside a workflow --
so each failed a required check and waited for a hand-applied label. The gate
now reads the whole documented list off the Conventional Commit type, with
tests for both directions (`feat(core): rework how deps are resolved` still
owes a fragment).

PR #1111, opened 2026-08-28
ci(core): fuzz the diff with ClusterFuzzLite

mapyr:
Draft while the acceptance test runs -- see the comments below.

## Why

`fuzz.yml` is a smoke test: 20 000 iterations from an empty corpus, so every
run re-searches the same shallow space. ClusterFuzzLite fuzzes what the pull
request **changed**, which is where a new defect is.

Refs #1104, #1101.

## What

- `.clusterfuzzlite/Dockerfile` + `build.sh`: `compile_python_fuzzer` over each
  `fuzz/fuzz_*.py`, with `fuzz/` on `PYTHONPATH` so the harnesses' import of
  `invariants` resolves at build time, plus the parse seed corpus as a
  `_seed_corpus.zip`.
- `.github/workflows/cfl-pr.yml`: build + `mode: code-change`, 180 seconds.
- The base image is digest-pinned with a dependabot entry for
  `/.clusterfuzzlite`, like every other Dockerfile here.

## Two things this PR is honest about

**Corpus persistence is not wired.** CFL carries a corpus between runs through
a separate `storage-repo`, which does not exist. Creating a repository is not
mine to do, so the workflow header says the gap out loud rather than letting
the file imply that runs accumulate. Each run still starts cold: this catches
what three minutes near the diff can reach -- more than the smoke job, less
than a campaign. That is the remaining half of #1104.

**Pinning the base image found a bug in my own ratchet from #1089.** The
dependabot-coverage test stripped a leading `.` from each Dockerfile's
directory, so `.clusterfuzzlite` was checked as `/clusterfuzzlite` -- a path
that matches no entry and no directory. Fixed here.

## How tested

- `pytest tests/ci` -- 79 passed. `shellcheck` and `actionlint` clean.
- The acceptance criterion is a planted crash: I will push one, show CFL
  catching it, and remove it before this leaves draft. Neither the build nor
  the fuzzing runs on my machine (it needs the OSS-Fuzz image), so this PR's
  own run is the first execution -- as with #1108, where the first run found a
  real bug in the harness.

## Agent metadata

- [x] LOC declared upfront: `~70`

### DIFF
--- a/.clusterfuzzlite/Dockerfile
+++ b/.clusterfuzzlite/Dockerfile
@@ -1,0 +1,11 @@
+# ClusterFuzzLite build image (#1104).
+#
+# `base-builder-python` is the image OSS-Fuzz uses for Python targets: it
+# carries `compile_python_fuzzer`, which wraps each harness with PyInstaller so
+# libFuzzer can execute it as a standalone binary. Building here rather than
+# installing atheris on a runner is the whole reason CFL can fuzz a diff.
+FROM gcr.io/oss-fuzz-base/base-builder-python@sha256:6b2b3a7e4a2da50de47f94925cffbe34747e1aae4f33d7f07ba6c681dd648b23
+
+COPY . $SRC/mcp-hangar
+WORKDIR $SRC/mcp-hangar
+COPY .clusterfuzzlite/build.sh $SRC/build.sh
--- a/.clusterfuzzlite/build.sh
+++ b/.clusterfuzzlite/build.sh
@@ -1,0 +1,19 @@
+#!/bin/bash -eu
+# Build the fuzz targets for ClusterFuzzLite (#1104).
+#
+# The harnesses import `invariants` from `fuzz/`, so that directory has to be
+# importable inside the built binary -- PyInstaller follows the import, but
+# only if it resolves at build time, which is what PYTHONPATH is for here.
+
+pip3 install --no-cache-dir .
+
+for harness in "$SRC"/mcp-hangar/fuzz/fuzz_*.py; do
+  PYTHONPATH="$SRC/mcp-hangar/fuzz" compile_python_fuzzer "$harness"
+done
+
+# Seed the parse target. The other two consume raw FuzzedDataProvider bytes, so
+# a hand-written seed for them would be an opaque blob -- libFuzzer builds and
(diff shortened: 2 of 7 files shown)


mapyr:
## The acceptance test passed, end to end

Planted a crash in `evaluate_tool` (`tool_name.startswith("aa")` ->
`RuntimeError`), pushed it, and both layers caught it. The commit is reverted;
both commits stay in the branch history as the evidence.

**ClusterFuzzLite** (`Fuzz the diff`, the criterion this issue asked for):

```
INFO - Trying to reproduce crash using: /tmp/.../crash-196135e0d17dbeda...
RuntimeError: planted crash for the ClusterFuzzLite acceptance test
==51== ERROR: libFuzzer: fuzz target exited
artifact_prefix='/tmp/.../'; Test unit written to /tmp/.../crash-196135e0...
INFO - Crash is reproducible.
```

All three targets built in the OSS-Fuzz image, `mode: code-change` picked the
target the diff touched, found it in under three minutes, reproduced it, and
failed the job.

**The plain atheris job** (#1108) found the same crash in **20 seconds** and
uploaded its own artifact. Two independent layers, both working.

A detail worth keeping: the CFL log also shows

```
argument_serialization_failed error=RecursionError: ... -- the call is refused uninspected
```

which is #1102's fix doing its job under fuzzing -- deep nesting now produces a
denial instead of an escaping exception, repeatedly, on inputs the fuzzer found
by itself.

## The corpus question answered itself

The header of `cfl-pr.yml` originally said corpus persistence needed a separate
`storage-repo` that nobody had created. The run disproved that:

```
WARNING - Could not find artifact: cifuzz-corpus-fuzz_policy_evaluate.
```

CFL's default filestore is **GitHub Actions artifacts**. Nothing was missing
except the job that produces them. So `cfl-batch.yml` is now in the PR: 30
minutes nightly in `batch` mode, uploading `cifuzz-corpus-<target>`, which the
PR job downloads. The two files are one feature -- a PR-only setup is a fuzzer
with amnesia, and I would have shipped exactly that on the strength of a guess
about how CFL stores corpora.

`actions: write` for the upload sits on the job rather than the workflow,
because the Token-Permissions ratchet from #1087 caught it at the top level.

Issue #1099, opened 2026-08-28
chore: promote the lab's seven-server reference pool into the repository


mapyr:
Part of #1094. The largest child, and optional if #1095 covers enough.

## Context

The local HA lab already runs **all seven** servers from
`modelcontextprotocol/servers` through the gateway -- the stdio ones behind
`mcp-proxy`, one pod each -- with per-server findings written up (tool counts,
what each needed, the egress policy `fetch` requires, the `mcp-proxy`
environment gotcha). That work exists outside version control, so it is one
laptop away from being lost, and nobody else can reproduce the pool.

It is also the only place where our claims about the official servers are
grounded in a real run rather than in an upstream README.

## Change

Promote it: manifests (or a compose file) that stand the seven servers up
against a Hangar, plus the notes on what each one needs. This is what
`scripts/containers.sh` was reaching for and never delivered (#1095).

Decide deliberately whether it lives under `examples/` (a user follows it) or
`tests/live` (CI or a human runs it against a real cluster). The lab's value so
far has been as an instrument that finds real defects -- it found the front_door
projection wipe and the session-ownership gap -- which argues for the second.

## Acceptance

- [ ] the seven servers stand up from files in this repository, on a stated
      runtime (kind, compose)
- [ ] each server's tool count and requirements recorded from a real run
- [ ] the stdio-via-`mcp-proxy` recipe is written down, including the
      environment-passing gotcha

mapyr:
Closing as **wontfix for the manifests, done for the knowledge**, with the
reasoning rather than a shrug.

## What was actually at risk

The issue's worry was that the pool "is one laptop away from being lost". Two
different things were: the *manifests* (a Dockerfile building all seven from a
pinned upstream commit, six Deployments, CRs, an egress policy) and the
*findings* -- what it took to make them work. Only the second is irreplaceable.
The manifests can be rewritten in an afternoon by anyone who has the findings;
the findings took a day of a cluster misbehaving.

## The knowledge is now in the docs

`guides/OFFICIAL_SERVERS.md` (mcp-hangar/docs#279) carries the bridge recipe
and the three pins that make it work:

- `mcp-proxy` >= 0.12.0 -- earlier lines are SSE-only, 0.12.0 is the first that
  mounts Streamable HTTP at `/mcp`.
- `mcp==1.28.1` in that image -- the reference servers are still SDK 1.x, and
[...]

Issue #1094, opened 2026-08-28
epic: run the official modelcontextprotocol/servers instead of our own demo providers


mapyr:
Replace the bespoke demo/provider layer with the servers from
[modelcontextprotocol/servers](https://github.com/modelcontextprotocol/servers),
so that what we ship as "here is a provider to point Hangar at" is the same
thing every MCP user already has.

## What the survey found

The answer is yes, and most of it is deletion rather than migration.

**1. The five `docker/Dockerfile.*` files have no builder.** Nothing in the
repository, the Makefile, CI or the docs builds them. `scripts/containers.sh`
runs `localhost/mcp-{memory,math,fetch}:latest`, images no documented command
produces. They are a vendoring layer that has never been exercised -- the same
advertised-but-unwired shape as #596 and #888.

**2. Two of them already wrap the official packages, one wraps a fork, one has
no upstream left.**

| File | Wraps | Official? |
|------|-------|-----------|
| `Dockerfile.filesystem` | `@modelcontextprotocol/server-filesystem` | yes, repackaged by us |
| `Dockerfile.memory` | `@modelcontextprotocol/server-memory` | yes, repackaged by us |
| `Dockerfile.fetch` | `@kazuph/mcp-fetch` | **no** -- a third-party fork; the official one is `mcp-server-fetch` on PyPI |
| `Dockerfile.sqlite` | `mcp-server-sqlite` (npm, third-party) | **no upstream** -- the official sqlite server is archived; PyPI `mcp-server-sqlite` last released 2025-04-25 |
| `Dockerfile.math` | our own `mock_provider.py` | n/a -- a test double |

**3. Prefer the packages over the published images.** The npm and PyPI
artifacts are current (`server-everything` 2026.8.18, `server-filesystem`
2026.7.10, `mcp-server-fetch` 2026.8.18). The `mcp/*` Docker Hub images are 12
to 16 months stale -- `mcp/everything` and `mcp/memory` were last pushed
2025-05-02, `mcp/filesystem` 2025-06-16. Only `mcp/fetch` is recent. So
`mode: subprocess` with `npx -y` / `uvx` is both the smaller change and the
fresher artifact, and it is already what the docs recommend.

**4. The docs are ahead of the repository.** `docs/getting-started/quickstart.md`
already runs `npx -y @modelcontextprotocol/server-filesystem`. Meanwhile
`examples/quickstart/config.yaml` -- the example in *this* repo, the one
[...]

PR #1119, opened 2026-08-28
chore(core): delete the demo images nothing built, and point the pool at official ones

### LINKED ISSUE #1095: chore: delete the unbuilt docker/Dockerfile.* layer and rebuild the pool on official servers
Part of #1094.

## Context

`docker/Dockerfile.{fetch,filesystem,memory,sqlite,math}` are built by nothing:
not the Makefile, not CI, not a script, not the docs. `scripts/containers.sh`
runs `localhost/mcp-memory:latest`, `localhost/mcp-math:latest` and
`localhost/mcp-fetch:latest` -- images no documented command produces, so the
script cannot work on a fresh checkout.

Two of the five repackage an official npm package; `fetch` repackages a
third-party fork (`@kazuph/mcp-fetch`) rather than the official
`mcp-server-fetch`; `sqlite` repackages a third-party npm whose upstream is
archived. See #1094 for the table.

## Change

Delete the five Dockerfiles and rebuild `scripts/containers.sh` (or replace it)
on the official servers, run as `npx -y @modelcontextprotocol/server-*` /
`uvx mcp-server-*`. The packages are current; the `mcp/*` Docker Hub images are
12-16 months stale, so images are the worse artifact here.

Open question for the PR to settle: whether `scripts/containers.sh` survives at
all. It exists to populate a container-discovery pool, which is exactly what
#C5 does properly. If C5 lands, this script is dead weight.

`sqlite` has no maintained upstream. Either drop it from the pool or keep it
with a comment saying it is third-party and unmaintained -- silently keeping a
third-party server in a repository whose thesis is provenance is the one option
that is wrong.

## Acceptance

- [ ] `docker/Dockerfile.*` removed, or each remaining one has a builder and a
      reason
- [ ] no reference in the repository to an image nothing builds
      (`grep -rn 'localhost/mcp-' .`)
- [ ] the base-image and dependabot ratchets from #1083 still pass (fewer
      Dockerfiles is fine; a Dockerfile with no dependabot entry is not)
- [ ] `fetch` is the official `mcp-server-fetch`, or gone

mapyr:
## Why

`docker/Dockerfile.{fetch,filesystem,memory,sqlite,math}` were built by nothing
in this repository -- not the Makefile, not CI, not a script. `scripts/containers.sh`
ran `localhost/mcp-*:latest`, images no command produced, and sent every
`podman run`'s stderr to `/dev/null`, so a missing image failed **silently**.
The script had never worked on a fresh checkout and said so to nobody.

Closes #1095. Refs #1094.

## What

- The five Dockerfiles are gone; `docker/` with them.
- `scripts/containers.sh` runs `mcp/filesystem`, `mcp/memory` and `mcp/fetch` --
  the official servers, published by Docker from `modelcontextprotocol/servers` --
  and **pulls up front**, so a pull failure is the operator's answer instead of
  five silent no-ops. `sqlite` is dropped: its upstream is archived and there is
  no maintained official replacement, and keeping a third-party server unmarked
  in a repository about provenance is the one option that is wrong.
- Dependabot's `/docker` entry removed with the directory; the Dockerfile-count
  floor in the ratchet lowered from 9 to 4, with a comment that it is a
  glob guard rather than a target.

## A second broken thing, found on the way

`examples/discovery/docker-compose.yml` named
`ghcr.io/modelcontextprotocol/server-fetch:latest` and `-server-memory:latest`.
**Those images do not exist.** Anonymous manifest requests return 403 where a
real public package answers 200 -- checked against
`ghcr.io/mcp-hangar/mcp-hangar` to make sure the probe itself was sound. It
also declared `mode: http` with `port: 3000` for servers that speak stdio. The
file could not have worked as written, in two independent ways. Now the same
official images with `mode: container`.

## One correction to the issue

#1095 says nothing builds those Dockerfiles. True of this repository; **not**
true of the docs, where `guides/CONTAINERS.md` gives the exact `podman build -f
docker/Dockerfile.sqlite` commands. So this PR breaks a documented guide, and
the companion docs PR is linked below rather than left for someone to find.

## How tested

- `pytest tests/unit tests/ci` -- 6788 + 76 passed. `shellcheck` clean on the
  rewritten script.
- Image existence verified against the registries (Docker Hub 200 for all
  three; the ghcr names 403).
- **Not** verified: that the official images run correctly under our discovery
  labels. Nothing in CI builds or runs this pool -- that is what #1099 is for --
  so this PR moves the script from "references images that cannot exist" to
  "references images that do", which is a smaller claim than "works".

## Risk and rollback

Anyone who built the old images by hand loses them; the changelog fragment says
so and points at the replacements. Revert the single commit.

## Agent metadata

- [x] LOC declared upfront: `~120` net deletion


mapyr:
`lint` failed, and the cause is a side effect of the deletion worth recording.

```
docker_source.py:47:  error: Unused "type: ignore" comment  [unused-ignore]
docker_source.py:164: error: Unused "type: ignore" comment  [unused-ignore]
docker_source.py:167: error: Unused "type: ignore" comment  [unused-ignore]
```

The root `docker/` directory -- five Dockerfiles and nothing else -- was
**shadowing the `docker` PyPI package for mypy**. `import docker` resolved to
that directory, so the SDK's real types were never seen, and three
`# type: ignore` comments were added to silence the resulting errors. One of
them even says "docker SDK has no type stubs", which was never true; it was
true of the directory.

Reproduced both ways on this branch:

```
$ mkdir docker && touch docker/.keep && mypy src | grep -c docker_source
0
$ rm -rf docker && mypy src | grep -c docker_source
3
```

So a directory of demo images nobody built had been quietly degrading type
checking of the real Docker discovery source. The ignores are removed; `mypy
src` is back to the 10 pre-existing errors in `tracing.py`, `langfuse.py` and
`redis_cache.py`, none of them new.

PR #1120, opened 2026-08-28
chore(core): the quickstart runs a provider that exists

### LINKED ISSUE #1096: chore: make examples/quickstart runnable against official servers
Part of #1094.

## Context

`examples/quickstart/config.yaml` declares:

```yaml
mcp_servers:
  openai-proxy:
    mode: docker
    image: my-org/openai-mcp:latest
    command: ["python", "-m", "openai_mcp_server"]
```

There is no such image and no such module. This is the example the README
points at, and `examples-compose.yml` brings it up in CI -- which passes,
because the compose file never actually starts that provider.

Meanwhile `docs/getting-started/quickstart.md` already does the right thing
(`npx -y @modelcontextprotocol/server-filesystem`), so the docs and the
repository disagree about what a quickstart looks like.

## Change

Point the quickstart at two official servers -- one with no side effects
(`server-everything` or `time`) and one that is obviously useful
(`server-filesystem` over a sandbox path) -- via `mode: subprocess` with
`npx -y` / `uvx`. Keep the capability block: it is the interesting part of the
example and it should describe a provider that exists.

Then make CI mean something: `examples-compose` should assert the gateway
projects a tool from the official server, not merely that the container is
HEALTHY.

## Acceptance

- [ ] a reader can `docker compose up` and call a real tool with no
      hand-editing and no credentials
- [ ] the capability block describes the provider actually declared
- [ ] `examples-compose` asserts a projected tool, and fails if the projection
      is empty
- [ ] no fictional image or module remains under `examples/`

mapyr:
## Why

`examples/quickstart/config.yaml` declared

```yaml
image: my-org/openai-mcp:latest
command: ["python", "-m", "openai_mcp_server"]
```

Neither the image nor the module exists. This is the example the README points
at, and the `examples-compose` job passed it every time -- it asserted
**healthy** and **config loaded**, both of which were true of a gateway serving
a provider that could never start.

Closes #1096. Refs #1094.

## What

- The provider is the official filesystem server from
  `modelcontextprotocol/servers`, as the image Docker publishes. No build, no
  account, no credentials.
- The capability block is rewritten to describe *that* provider -- `egress: []`
  and `dns_allowed: false` for a server with nobody to talk to, matching the
  runtime `network: none` beside it. A capability block describing a different
  provider than the one declared is worse than none.
- `examples-compose` now sends `tools/list` and **fails on an empty
  projection**, with the gateway log dumped on failure.

## The socket, said out loud

Container mode means the gateway starts the provider through the Docker API, so
the compose file mounts `/var/run/docker.sock` and adds the gateway to the
socket's group -- the image runs as the unprivileged `hangar` user, which is
otherwise not allowed to read it. **That is equivalent to root on the host**,
and the README says so rather than leaving it as an unremarked line in a
volumes list. A subprocess or remote provider needs none of it.

The gid is not a constant (999 on the runners, elsewhere not), so it is
`${DOCKER_GID:-999}` with the override documented, and CI computes it with
`stat -c %g`.

## What I could not verify locally

No container runtime on this machine (podman is installed but its VM has never
been started), so **this PR's own `examples-compose` run is the first execution**
of the configuration. The two things it could plausibly get wrong are the
socket group and the sandbox directory's ownership; both now have explicit
steps, and both fail loudly rather than silently if I got them wrong. If the
run is red, the fix is in the same place as the evidence.

A subprocess provider (`npx -y @modelcontextprotocol/server-filesystem`) would
have avoided the socket entirely and is what the docs recommend for that shape
-- it is not usable here because the published image is `python:3.14-slim` with
no node and no uv, so there is nothing for `npx` to be.

## How tested

- `pytest tests/ci` -- 76 passed. `actionlint` clean.
- Image existence checked against Docker Hub.
- Everything else: this PR's run.

## Agent metadata

- [x] LOC declared upfront: `~120`

### DIFF
--- a/.github/workflows/examples-compose.yml
+++ b/.github/workflows/examples-compose.yml
@@ -73,6 +73,18 @@ jobs:
       - name: Build the image under test
         run: docker build -t "$HANGAR_IMAGE" .
 
+      - name: Prepare the host paths the example mounts
+        run: |
+          set -euo pipefail
+          # The quickstart's provider is given a sandbox directory; compose
+          # would otherwise create it root-owned and the provider could not
+          # write to it.
+          sudo mkdir -p /srv/hangar-quickstart/data
+          sudo chmod 777 /srv/hangar-quickstart/data
+          # The gateway runs unprivileged and reaches the Docker socket through
+          # a supplementary group, whose gid is the host's, not a constant.
+          echo "DOCKER_GID=$(stat -c %g /var/run/docker.sock)" >> "$GITHUB_ENV"
+
       - name: Bring ${{ matrix.example }} up and wait for healthy
         working-directory: examples/${{ matrix.example }}
         run: |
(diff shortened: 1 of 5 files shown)


mapyr:
`compose-up (quickstart)` failed on the **new** assertion -- the check doing
its job, and then two bugs in the check itself.

Healthy: yes. Config loaded: yes. Provider reachable: `HTTP 400`.

I reproduced it against a local gateway instead of guessing, which took about a
minute and answered both questions exactly:

```
400  params._meta is missing the required envelope key(s):
     io.modelcontextprotocol/clientCapabilities
```

The 2026-07-28 envelope needs **both** keys; I sent only `protocolVersion`.
Fixing that produced the second:

```
-32020  mcp-method header does not match the request body's method
```

SEP-2243 wants an `Mcp-Method` header agreeing with the body. With both fixed
the same request returns a tool list.

**And `tools/list` was the wrong question.** The quickstart uses the default
topology, where `tools/list` returns the gateway's own `hangar_*` meta-API --
present and complete whether or not a single provider works. My assertion would
have gone green against exactly the broken configuration it exists to catch:
the same blind spot as `healthy`, moved one step along.

So it now **calls** the provider: `hangar_call` -> `filesystem.list_directory`
on `/data`, which exercises config, the Docker socket, the container start and
the invocation end to end. The README carries the corrected request and names
the three things that answer 400 rather than a wrong result.

Both diagnostics also improved: the probe now prints the HTTP error **body**,
not just the status. The first run told me "HTTP Error 400: Bad Request" and
nothing else, which is why it needed a local reproduction at all.

Issue #1184, opened 2026-09-02
release: re-fired checks can land in action_required, which a workflow cannot clear


mapyr:
**Severity:** p2 — the automated recovery added by #1181 cannot finish the job on its own; a release still stops until a human notices
**Seen on:** #1173 (release 2.17.1), 2026-09-02
**Follows:** #1180 / #1181

## What #1181 does, and where it stops

#1181 made the changelog assembly detect that it pushed with `GITHUB_TOKEN` — a credential that does not trigger workflows — and recover by closing and reopening the release PR, which fires `pull_request: reopened` against the head already pushed.

That works, and it was verified by hand on #1173. But the runs it fires can land in `action_required` rather than starting, and then the PR is blocked exactly as before:

```
gh run list --branch release-please--branches--main--components--mcp-hangar
2026-09-02T21:28:01Z 2a634e87 pr-body: completed/action_required
2026-09-02T21:28:01Z 2a634e87 pr-title: completed/action_required
2026-09-02T21:28:00Z 2a634e87 Security: completed/action_required
... 10 runs in total
```

`gh pr checks` reports "no checks reported", the required contexts stay "expected", and the merge is refused with the same `13 of 13 required status checks are expected` that started this whole thread. Clearing it took ten `POST /actions/runs/<id>/approve` calls.

**A workflow cannot approve its own runs.** `/approve` needs a credential belonging to someone with the right to approve, and the run doing the recovery does not have one — `GITHUB_TOKEN` cannot, and handing it a PAT that can would recreate the "gate as theatre plus a long-lived secret" problem the registry-publish gate deliberately avoids. So the recovery in #1181 closes half the hole: it starts the runs, and something else has to release them.

## Why the runs need approval at all

Not established, and worth finding out before choosing a fix. Candidates:

- an Actions setting requiring approval for runs from certain actors, which the release bot app hits and a human push does not;
- the same fallback as #1180: with the app token missing, the head commit is authored by `github-actions[bot]`, and the repo may treat that actor's runs as needing approval.

The second would make this a *consequence* of #1180 rather than an independent problem, which would be good news — fixing the app token would fix both. Worth checking `gh api repos/mcp-hangar/mcp-hangar/actions/permissions` and the "Approval for running fork/actor workflows" setting before writing any code.

## What would actually close it

In rough order of preference:

[...]

Issue #1180, opened 2026-09-02
release: changelog assembly can push with GITHUB_TOKEN, leaving a release PR whose required checks never run


mapyr:
**Severity:** p2 — the release PR cannot be merged at all, including with `--admin`, and the failure looks like a branch-protection problem rather than a workflow one
**Seen on:** #1173 (release 2.17.1), 2026-09-02

## What happened

`gh pr merge 1173 --squash --admin` was refused:

```
GraphQL: Repository rule violations found
13 of 13 required status checks are expected.
```

Not *failing* — **expected**. `gh pr checks 1173` reported "no checks reported on the 'release-please--branches--main--components--mcp-hangar' branch": the head commit had no workflow runs at all, so every required check stayed pending forever and the ruleset blocked the merge. `--admin` does not help, because the rule is a ruleset requirement rather than classic branch protection.

## Cause

The two commits on the release branch were pushed by different identities:

```
c42049de github-actions[bot]         chore(release): assemble changelog and upgrade notes for 2.17.1
e340abda mcp-hangar-release-bot[bot] chore(release): release 2.17.1
```

`release-please.yml` pushes the assembled changelog with:

```yaml
PUSH_TOKEN: ${{ steps.app-token.outputs.token || secrets.GITHUB_TOKEN }}
```

On this run the app-token step produced nothing, so the push fell back to `secrets.GITHUB_TOKEN` — and **a push made with `GITHUB_TOKEN` does not trigger workflows**, by design, to prevent recursive runs. The release-please commit underneath it was pushed by the app and would have triggered them; the assembly commit landed on top and became the head with no runs of its own.

The previous release PR (#1157) carried 32 checks and merged normally, so this is intermittent: it happens exactly when the app-token step fails and the fallback is taken.

## Why the fallback is worse than a failure

The fallback exists so the job does not die without the app token. What it actually produces is a release PR that looks complete — correct notes, correct version bump, green-looking history — and can never be merged by anyone, with an error message that points at branch protection. Working that out took a while; the next person pays the same cost.

## Workaround (used here)

Close and reopen the PR. That fires `pull_request: reopened` on the existing head, the workflows run, and the branch becomes mergeable. No history is rewritten.

## Fix

1. Fail the assembly step when `steps.app-token.outputs.token` is empty rather than falling back to `GITHUB_TOKEN`. A release PR without notes is recoverable; one that cannot be merged is not, and the fallback is not actually a degraded mode — it is a broken one.
[...]

PR #1185, opened 2026-09-02
fix(ci): make the release PR's checks start on their own

mapyr:
## Why

Correcting #1181, which I shipped on an inference that the evidence does not support.

#1181 recovered when `PUSH_TOKEN` equalled `GH_TOKEN`, reasoning that a push made with the built-in `GITHUB_TOKEN` does not trigger workflows. The reasoning is sound; the condition was still wrong for the case it was written for. On that release run the app-token step executed with `app-id`/`private-key` present, and the release commit is authored by `mcp-hangar-release-bot[bot]` — so the app token was available, `PUSH_TOKEN` was not `GH_TOKEN`, and **the recovery would never have fired**. It only unblocked #1173 because I ran the close/reopen by hand.

What the API actually shows on the two heads of that PR:

```
c42049de  10 runs, actor=github-actions[bot], conclusion=failure, zero jobs
2a634e87  10 runs, conclusion=action_required
```

Both produce **no check runs**, which is what the merge gate reads and reports as `13 of 13 required status checks are expected`. Re-firing the same ten workflows as a human actor produced checks immediately — at 21:17 on the first head (my close/reopen) and after `POST /actions/runs/{id}/approve` on the second.

The cause of the bot-actor gating is still open (#1184, and the app-token question on #1180). This PR does not depend on it.

## How

The condition is now the symptom, checked after the push: **if the head has no check runs, reopen the PR to fire them.** That is true whatever the underlying reason turns out to be, and it is exactly the state that blocks the merge. It reads `/commits/{sha}/check-runs`, not `/actions/runs` — the two differ precisely where it matters, since ten workflow runs contributing zero checks is the shape that happened.

It also states its own ceiling instead of pretending to be complete: the reopen uses this job's credential, and a run whose actor is `github-actions[bot]` is what produced no checks in the first place, so the retry can come back empty too. When it does, the job now fails with `::error::` carrying the exact `/approve` command, rather than leaving a release PR that looks finished and cannot be merged. A stall that explains itself costs a minute; this one cost an hour.

## How tested

`tests/ci/test_release_pr_checks_are_refired.py` — the four stub-`gh` tests of the recovery script are unchanged and still pass. The two assertions about the assembler are rewritten: the token comparison must be gone, the check-runs lookup must be there, and the error path must name the `/approve` call rather than describe it.

`tests/ci`: 83 passed. `bash -n` clean, `ruff` clean.

The grace period is `CHECK_GRACE_S` (45s default), so a test or a manual run can drive it without waiting.

https://claude.ai/code/session_01N2Hz4XYkz5QLipERGBE4cL

### DIFF
--- a/scripts/assemble_release_changelog.sh
+++ b/scripts/assemble_release_changelog.sh
@@ -115,25 +115,60 @@ git -c user.name="github-actions[bot]" \
 # nothing is wrong. The app token does trigger them.
 git push "https://x-access-token:${PUSH_TOKEN}@github.com/${REPO}.git" "HEAD:${branch}"
 
-# Fix the fallback rather than describe it.
+# Recover on what is observable, not on a theory about why.
 #
-# The condition used to be `PUSH_TOKEN = GH_TOKEN && RELEASE_BOT_APP_ID unset`,
-# on the assumption that the fallback only happens where no app is configured.
-# It also happens where the app IS configured and the token step yields nothing
-# -- an expired installation, a permissions change, a transient failure -- and
-# that is the case that reached #1173: the app id was set, so the warning was
-# skipped, and the release PR sat with 13 required checks in "expected" that no
-# push would ever satisfy. `--admin` does not clear a ruleset requirement, so
-# the release was stuck behind an error message pointing at branch protection.
+# The first version of this guessed at the cause: it compared PUSH_TOKEN to
+# GH_TOKEN, on the reasoning that a push made with the built-in GITHUB_TOKEN
+# does not trigger workflows. That reasoning is sound and the condition was
+# still wrong -- on the run it was written for, the app token WAS available, so
+# the comparison was false and the recovery never ran. The release PR sat with
+# 13 required checks "expected" anyway (#1173).
 #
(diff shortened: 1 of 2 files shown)


mapyr:
Second commit added, and it changes what this PR is: the cause is now known, so this is no longer only a recovery.

The organisation requires approval to run workflows for **first-time contributors** — anyone with nothing merged into this repository — and the setting checks *the actor of the pull request event*, not only its author. This script signed its commit as `github-actions[bot]` while pushing with the release app's token, and `github-actions[bot]` has nothing merged here: the assembly commit lives only on the release branch, and the squash into `main` is attributed to the PR's author, the app. First-time contributor on every release, forever.

That also explains dependabot being unaffected (it has merged PRs) and close/reopen working (it made a member the actor).

So the fix is to commit as the identity that actually pushes: `mcp-hangar-release-bot[bot]`. The `github-actions[bot]` identity stays for the no-app-token path, where it really is who pushed.

The alternative was relaxing the org setting to "first-time contributors who are new to GitHub". I did not take it: this is a public repository, and that option lets a fork PR from any established account run workflows unreviewed — a security control paying for a release-process bug.

Full detail on #1184.

Issue #1189, opened 2026-09-03
EPIC-FUNNEL: the first run ends in a verdict (2.18.0)


mapyr:
**Status:** proposed · **Target release:** 2.18.0 (minor: new config keys, no behavior change when absent)
**Scope rule:** nothing else ships in 2.18.0. Security fixes excepted. Every other open thread waits.

## Why

The website says "Every MCP tool call ends in a verdict." The first config Hangar writes for a new user ends every call in `allow`. This epic makes the first run produce a deny, on a laptop, without a cluster, in the client the user already has open (Claude Code / Cursor / Claude Desktop).

Distribution thesis this serves: the first 100 users are people with 5-15 MCP servers in `.mcp.json` who have read about tool poisoning and rug pulls and have no deterministic tool that stops it locally. The cluster story is for the next 1000 and is out of scope here.

## Evidence (verified on the 2.17.1 wheel from PyPI and live over stdio, 2026-09-03)

| # | Observation | Where |
|---|-------------|-------|
| E1 | `mcp-hangar init -y` writes `mcp_servers` (filesystem/fetch/memory) + `health_check` + `logging`. No `tool_access`, no `tool_projection.pins`, no `auth`. | `server/cli/commands/init.py`, generated config |
| E2 | `tool_access.mode: front_door` + auth off, caller over stdio → `tools/list` returns **0 tools**; log `empty_projection reason=no_identity ... Fail-closed deny`. front_door is unreachable over stdio. | `fastmcp_server/flat_tool_projection.py` |
| E3 | Default `egress` mode over stdio → 22 `hangar_*` tools, ~11k tokens of `tools/list` alone (chars/4, one upstream). The user's own tools are reachable only through `hangar_call`. | live probe |
| E4 | `tool_projection.pins: {fetch: "000..."}` + `hangar_call` → `error_type: ToolDigestMismatchError`, message `Tool 'fetch' schema does not match its pinned digest`. `digest_enforcement` defaults to `block` (`tool_projection.py:309`). Enforcement works; nothing leads a user to it. | `server/tools/batch/executor.py:1264-1302` |
| E5 | No command, tool or REST field exposes a computed digest. `compute_tool_digest` is a single public function (`domain/services/digest_computation.py:27`) used by the registry (`tool_projection.py:159`). | grep on wheel |
[...]

PR #1196, opened 2026-09-03
feat(core): give a stdio session the identity its config declares

### LINKED ISSUE #1190: WS-0: ADR-026 -- stdio is an authenticated transport
Part of #1189 (EPIC-FUNNEL). **First workstream -- everything else in the epic depends on this one.**

## Decision

The process that spawns Hangar over stdio is the trust boundary: the OS user launched it. A config block declares who that caller is:

```yaml
auth:
  stdio:
    principal:
      id: local-user          # required
      tenant_id: local        # required
      roles: [viewer]         # default: read-only management surface
```

- When the serving transport is stdio and the block is present, the identity middleware sets the identity context for the session from the block. No credential is checked; none exists to check.
- Absent block → behavior identical to 2.17.1 (anonymous caller). This is what makes it a minor.
- HTTP transport ignores the block. `--unsafe-no-auth` semantics unchanged.
- `refuse_pins_that_no_caller_can_match` (#902) must treat per-tenant pins for the stdio principal's `tenant_id` as matchable.
- `empty_projection{reason=no_identity}` must not fire over stdio when the block is present.
- The management surface follows `roles` through the existing `TOOL_PERMISSIONS` table (#910/#912).

**Not** an alternative: projecting all-tenants flat tools to anonymous callers. That dissolves the fail-closed model #902 was built on.

## Default role: `viewer`

`ROLE_VIEWER` already exists (`src/mcp_hangar/auth/roles.py:216`), so no role is invented. Two things to verify rather than assume while implementing:

1. `viewer` holds `providers:read`, `provider:read`, `provider:list`, `tool:list`, `metrics:read`, `group:read`, `group:list`, `discovery:read` -- so the projected management surface is the fleet reads plus group list, metrics and discovery reads, not just `hangar_status`/`hangar_health`. Record the actual projected list and its token cost in the PR; if it is large enough to defeat the point of front_door, say so here before widening the epic.
2. `viewer` does NOT hold `tool:invoke`. The flat path authorizes upstream calls through the tool-access policy inside `BatchExecutor`, not through RBAC (`tool:invoke` is checked only by `hangar_call`'s `_authorize_calls`), so a flat call should still succeed. The e2e has to prove it.

## Acceptance

- ADR file `docs/adr/ADR-026-stdio-principal.md`.
- e2e over stdio with the SDK client (not curl):
  - front_door + block → `tools/list` returns the upstream's flat tool names, and a flat `tools/call` under `roles: [viewer]` succeeds;
  - no block → 0 tools and the `no_identity` warning (regression guard).

mapyr:
Closes #1190. WS-0 of #1189. Implements [ADR-026](https://github.com/mcp-hangar/docs/blob/main/adr/ADR-026-stdio-is-an-authenticated-transport.md) (merged as Proposed; flip to Accepted follows this).

## Why

`front_door` is the projection we recommend and it was unreachable over stdio. `IdentityMiddleware` returns early for any ASGI scope that is not `http`/`websocket`, and `run_stdio` serves a pipe — no scope at all. So the per-request flat map was built for `tenant_id = None`, the deny-all branch ran, and `tools/list` returned zero tools with `empty_projection reason=no_identity … Fail-closed deny`. That left `egress` as the only working local mode: 22 `hangar_*` tools, ~11k tokens of `tools/list` for one upstream, and the user's own tools reachable only through `hangar_call`.

ADR-026 settles the trust question: a stdio server is not listening on anything, it was spawned by a session the OS already authenticated, and a secret stored in the same file that declares the principal would authenticate the file rather than the caller. So `auth.stdio.principal` declares the caller and nothing is checked.

```yaml
auth:
  stdio:
    principal: {id: local-user, tenant_id: local, roles: [viewer]}
```

## What changed

- **`auth/stdio_principal.py`** — holds the declared principal for the process. A stdio server serves exactly one session over one pair of pipes, so there is no second caller for a per-request binding to tell apart.
- **`get_identity_context`** — falls back to the declaration when nothing bound the contextvar. One chokepoint rather than a fallback at each reader, so the flat projection, the batch executor and the pin scope all see it; and a bound context still wins, so an HTTP request is untouched.
- **`management_tools_for`** — a declared principal answers first (on stdio there is nothing else to ask), and its roles are resolved against the same `TOOL_PERMISSIONS` entries the authorizer uses. This can only *narrow*: with auth off `authorize_tool` already permits every management tool while the front door showed none, so a declaration moves that list from empty to a subset. Only built-in roles resolve; a role defined in the role store belongs to a deployment that has auth on, and that deployment takes the existing path.
- **`refuse_pins_that_no_caller_can_match`** (#902) — stops refusing per-tenant pins addressed to the declared tenant, since a caller can now carry it. Pins for every other tenant still refuse. It reads the bound principal rather than the raw document, so an HTTP run (where the block is ignored) still refuses them.
- **`bootstrap(..., stdio=)`** — the transport is threaded in from `run_server`, because the block must be read only on stdio, and read before the pin refusal answers its question.
- **`config_schema`** — `stdio` added to the closed `auth` key set, or a config carrying it is rejected as a typo.

## How tested

`6980 passed` (full `tests/unit`), `ruff format --check` and `ruff check` clean, `mypy` unchanged from `main` (same 10 pre-existing errors in 3 files, none in the touched ones).

New unit file, 17 tests, covering parse (including a half-written principal being dropped rather than guessed at), stdio-only application, the identity fallback and the bound-context precedence, the role-derived surface, and both directions of the pin refusal. The transport gate was probed rather than trusted: forcing `_declare_stdio_principal` to declare unconditionally turns `test_http_ignores_the_block` red, and reverting restores green.

**Live over stdio with the SDK client** (`ClientSession` + `stdio_client` against `mcp-hangar serve`), one tiny upstream exposing `echo`:

| Config | `tools/list` | flat `tools/call` |
| ------- | ------------- | ------------------ |
| block present | `['echo', 'hangar_details', 'hangar_discovered', 'hangar_group_list', 'hangar_health', 'hangar_list', 'hangar_metrics', 'hangar_sources', 'hangar_status', 'hangar_tools']` | `echo` → `hi`, `isError` unset |
| block absent | `[]` + `empty_projection reason=no_identity` | — |

Two things that probe confirms beyond the acceptance criteria: the projected management surface under `roles: [viewer]` is exactly the nine read-only tools (no `hangar_start`/`stop`/`load`/`unload`/`reload_config`), and a flat call succeeds under `viewer` even though that role has no `tool:invoke` — the flat path authorizes through the tool-access policy in `BatchExecutor`, while `tool:invoke` gates `hangar_call`. That was an assumption in #1190; it is now measured.

## Not in this PR

The config reference in `mcp-hangar/docs` cannot document `auth.stdio` until this merges — the docs config gate validates every YAML block against the schema in the core checkout, so the page would fail its own CI today. It follows as a docs PR once this is on `main`.

### DIFF
--- a/changelog.d/1190-stdio-principal.added.md
+++ b/changelog.d/1190-stdio-principal.added.md
@@ -1,0 +1,21 @@
+**core:** a stdio session can now be given an identity, so `tool_access.mode: front_door`
+is reachable without a cluster. Declare the caller the spawning process implies:
+
+```yaml
+auth:
+  stdio:
+    principal:
+      id: local-user
+      tenant_id: local
+      roles: [viewer]      # default; read-only
+```
+
+With the block present and the transport stdio, `tools/list` serves the upstream's own
+flat tool names instead of the fail-closed empty list, per-tenant digest pins addressed to
+that tenant are matchable (and no longer refused at boot), and the `hangar_*` management
+surface follows the declared roles -- `viewer` shows the fleet reads and nothing that can
+change state. No credential is checked, because a stdio server is not listening on
+anything: the OS user who launched the process is the trust boundary (ADR-026).
+
+Absent the block, nothing changes: the caller is anonymous and the front door stays empty.
+HTTP ignores the block entirely and keeps using its credential channel.
--- a/src/mcp_hangar/auth/config.py
+++ b/src/mcp_hangar/auth/config.py
@@ -8,6 +8,10 @@ import os
 from dataclasses import dataclass, field
 from typing import Any
 
+from mcp_hangar.logging_config import get_logger
+
+logger = get_logger(__name__)
+
 
 @dataclass
 class ApiKeyAuthConfig:
@@ -262,6 +266,23 @@ class RateLimitConfig:
 
 
 @dataclass
+class StdioPrincipalConfig:
+    """The caller a stdio session is declared to be (ADR-026).
+
+    Attributes:
+        id: Principal id, e.g. "local-user".
+        tenant_id: Tenant the caller belongs to; scopes policy and pins.
+        roles: Role names resolved through the ordinary RBAC path. Defaults to
+            read-only ``viewer`` so a first run can ask the gateway what it
+            thinks is happening without being able to change anything.
+    """
+
+    id: str
+    tenant_id: str
+    roles: list[str] = field(default_factory=lambda: ["viewer"])
+
+
+@dataclass
 class AuthConfig:
     """Authentication and authorization configuration.
 
@@ -292,6 +313,10 @@ class AuthConfig:
     opa: OPAConfig = field(default_factory=OPAConfig)
 
     role_assignments: list[RoleAssignment] = field(default_factory=list)
+
+    #: Declared principal for a stdio session (ADR-026). Ignored over HTTP,
+    #: which has a credential channel of its own.
+    stdio: StdioPrincipalConfig | None = None
 
 
 def _parse_tenant_audiences(raw: Any) -> dict[str, str]:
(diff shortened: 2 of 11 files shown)


mapyr:
Second commit: `lint-imports` refused the first shape — `mcp_hangar.context` is the shared kernel and reaching `auth.stdio_principal` to answer `get_identity_context` pulls the domain in behind it (`Contracts: 0 kept, 1 broken`). Inverted: the kernel owns the slot and the read (`set_fallback_identity`), the auth layer writes it. `get_stdio_identity` became unreferenced and the dead-symbol gate caught it, so it is deleted rather than left public.

Re-verified after the refactor: `6980 passed`, contracts `1 kept, 0 broken`, and both live stdio probes unchanged — `echo` projected and callable with the block, `[]` without it.

PR #1198, opened 2026-09-03
feat(core): make `init` write a config that ends a call in a verdict

### LINKED ISSUE #1192: WS-2: `init` writes a verdict-producing config
Part of #1189 (EPIC-FUNNEL). Depends on 1191.

Generated config (defaults, `-y`):

```yaml
mcp_servers:
  <selected servers as today>
    tool_projection:
      digest_enforcement: block       # explicit, even though it is the default
      pins:
        <tool>: <sha256>              # taken during the smoke test -- servers are up anyway
tool_access:
  mode: front_door
auth:
  stdio:
    principal: {id: local-user, tenant_id: local, roles: [viewer]}
health_check: {enabled: true, interval_s: 30}
logging: {level: INFO, json_format: false}
```

- Pins are captured **inside the smoke test step** (servers are already started there). `--skip-test` therefore also means `--skip-pins`; the panel must say so instead of "All passed" (fixes E6: the panel reflects what actually ran, rather than `smoke_test_passed: bool = True` defaulting to a pass, `init.py:239`).
- Client detection, a `--client` flag, all three writable:
  - Claude Desktop -- existing (`claude_desktop_config.json`).
  - Claude Code -- user scope `~/.claude.json` (`mcpServers`), project scope `./.mcp.json`.
  - Cursor -- `~/.cursor/mcp.json`, project `./.cursor/mcp.json`.
  - Verify the exact keys against each client's current docs at implementation time; do not trust this table from memory. Multiple clients detected → ask, or `--client all`.
- Fossils (E8): step numbering (currently 0, 1, 2, 4); "McpServer(s)"/"mcp_server(s)" in human-facing strings → "MCP server(s)"; generated-config header link → `https://mcp-hangar.io/docs/reference/configuration` (the current one does not resolve).

## Acceptance

`init -y --config-path X` produces a config that passes `mcp-hangar config check`, contains pins for every tool the smoke test saw, and serves flat tools over stdio (reuse the WS-0 e2e).

mapyr:
Closes #1192. WS-2 of #1189. **Stacked on #1197** (`feat/pin-command`) — merge that first; this PR targets it, so its own diff stays reviewable.

## Why

`init -y` wrote `mcp_servers`, `health_check` and `logging`. No `tool_access`, no pins, no identity. So a first run served 22 `hangar_*` tools to an anonymous caller and ended every tool call in `allow`, while the site says every MCP tool call ends in a verdict. This makes the generated file govern:

```yaml
mcp_servers:
  fetch:
    ...
    tool_projection:
      digest_enforcement: block
      pins: {fetch: fe1e2fda…}
tool_access:
  mode: front_door
auth:
  stdio:
    principal: {id: local-user, tenant_id: local, roles: [viewer]}
```

Pins are captured **inside the smoke-test step**, where every server is already up, so nothing is started twice. `--skip-test` therefore writes no pins — an unverified pin would refuse every call to a tool nobody digested — and the panel says so instead of the old `smoke_test_passed: bool = True` default printing "All passed" for a test that never ran (E6).

## Clients

`init` knew Claude Desktop only. It now writes, detected or named with `--client`:

| client | user scope | project scope |
| ------- | ----------- | -------------- |
| Claude Code | `~/.claude.json` | `./.mcp.json` |
| Cursor | `~/.cursor/mcp.json` | `./.cursor/mcp.json` |
| Claude Desktop | `claude_desktop_config.json` | — |

Paths and keys were read off each client's current docs at implementation time, as #1192 asked — Claude Code declares `"type": "stdio"`, Cursor infers it from `command`. `--skip-claude` still works, as an alias of `--skip-clients`.

The entry is **merged** into the client's `mcpServers`. The old Desktop writer assigned the whole map (`new_config["mcpServers"] = {"mcp-hangar": …}`), so a user with eight servers configured lost seven of them the moment the wizard ran. `ClaudeDesktopManager` is deleted rather than left beside the new writer — two things that know where Desktop keeps its config is how they drift.

## Two defects found while making the first run actually work

**1. The default bundle could not start.** `uvx_package` was derived from the npm name for every server, and the wizard prefers uvx ("dogfooding — our tool is Python-based"). Seven of ten names were wrong:

| name | what it is |
| ----- | ----------- |
| `mcp-server-filesystem`, `mcp-server-github`, `mcp-server-slack`, `mcp-server-google-maps` | not on PyPI |
| `mcp-server-memory` | exists, ships no executable |
| `mcp-server-postgres` | someone's personal package, summary "Add your description here" |
| `mcp-server-brave-search` | an Amazon placeholder, "This package is reserved" |
| `mcp-server-sqlite` | resolves, but no author and no repository link |

The first five meant two of the three servers in the `starter` bundle died on first run (`reader_died`, `does not provide any executables`). The rest are worse than broken: uvx would fetch and run a stranger's code under an official-looking name, from a tool whose pitch is supply-chain governance. Only `mcp-server-fetch` and `mcp-server-git` are published by Anthropic from `modelcontextprotocol/servers` (checked via `pypi.org/pypi/<name>/json`); those two keep their `uvx_package` and the rest use their real npm packages. The registry comment records the check so the next addition is deliberate.

The unit tests here asserted the bug's own shape — "all starter providers should have uvx_package", "names follow the mcp-server-* convention" — and are rewritten to pin the rule that replaced it.

**2. `health_check:` has never been read.** The generated config wrote it; the worker takes its interval from `HEALTH_CHECK_INTERVAL_SECONDS`, and `health_check` is not among the sections `config_schema` knows — so every generated config logged `unknown_config_key`, and an operator tuning `interval_s` tuned nothing. Removed rather than wired: wiring it is behaviour, and 2.18.0 is closed to anything else. Worth a follow-up issue if the knob should exist.

## How tested

`7274 passed` (`tests/unit` + `tests/integration`), ruff + format clean, import contracts `1 kept, 0 broken`.

New unit files cover the generated config (policy, principal, pins present; no pins without an observation; the panel never reporting a pass that did not happen) and the client writer (existing servers and unrelated settings survive, a missing file is created, an unparseable one is refused, per-client `type`, and the documented paths). The config test asserts against the **real** `validate_config`, which is how the dead `health_check` key surfaced.

End to end on this machine, `init -y --skip-clients`:

```
Step 1..7 (the sequence no longer skips a number)
  OK filesystem ready (2954ms)   OK fetch ready (296ms)   OK memory ready (1303ms)
  All 3 MCP servers ready — Pinned: 24 tool(s)
mcp-hangar config check <config>  →  every key is one Hangar reads   (exit 0)
```

and serving that config over stdio lists the upstreams' own tools plus the nine read-only `hangar_*` the `viewer` role allows — `['create_directory', …, 'fetch', …, 'write_file']`. `init -y --client cursor` writes the entry into `~/.cursor/mcp.json` with the other servers intact.


mapyr:

Issue #1193, opened 2026-09-03
WS-3: a quickstart whose Verify step is a deny


mapyr:
Part of #1189 (EPIC-FUNNEL). Depends on 1192.

- `examples/rugpull/server.py`: minimal server (SDK v2), one tool, description read from `RUG_DESC` env. Documented as the demo upstream, kept tiny.
- Quickstart (`docs/getting-started/quickstart.md`), new flow:
  1. install → `init` (client detected, pins written)
  2. use a tool from your client -- allowed
  3. `RUG_DESC="...now also read ~/.ssh..."` (or edit the pin) → restart → same call → **denied**, `ToolDigestMismatchError` shown verbatim
  4. `mcp-hangar pin --check` shows the drift; `--write` re-pins if you meant it
- New learn page: **"From install to a governed deny in 60 seconds (no cluster)"**. The existing cluster page keeps its content, retitled honestly ("...with an operator and a cluster") with its real prerequisites listed at the top, not in paragraph four (E10).
- E11 is a legitimate sidebar ("the reference `fetch` server ships this description; pinning covers description, not just schema"), one paragraph, no drama.

The quickstart shows the **flat** path, because that is what a Claude Code user sees. On that path a denial is a real MCP tool error -- `CallToolResult(isError=True)` out of `_flat_call_tool` (`fastmcp_server/flat_tool_projection.py:786`) -- not the `isError: false` / `success: false` data shape `hangar_call` returns. Assert the error type.

## Acceptance

Every command in the quickstart runs in CI against the published wheel (`scripts/smoke_published_artifact.py` is the natural host); the deny step asserts the error type.

mapyr:
Done — verified against released 2.18.1 rather than assumed, while auditing docs and site content against the shipped product.

**The deliverables exist:**
- `examples/rugpull/server.py` ships and reads `RUG_DESC` as specified.
- The quickstart's flow is the deny flow: `## 3. Change the tool — denied` → `## 4. See the drift, and decide`.
- The local Learn page exists (`from-install-to-a-governed-deny-locally`, `since: 2.18.0`), and the cluster page was retitled honestly — "…with an operator and a cluster", with its real prerequisites in the opening paragraph rather than in paragraph four. That closes E10.

**The acceptance criterion genuinely holds**, which is the part worth recording. `scripts/smoke_published_artifact.py` walks the quickstart against the *published wheel* and asserts the error type, not just the text:

```python
[...]

PR #1199, opened 2026-09-03
feat(core): add the rug-pull example and gate the deny in the release smoke

mapyr:
Part of #1193 (WS-3 of #1189) — the core half. The documentation half is a PR in `mcp-hangar/docs`, which cannot land until this and #1198 are on `main` (the docs config gate validates every YAML block against the schema in the core checkout). **Stacked on #1198**, which is stacked on #1197; merge in that order.

## Why

WS-3's acceptance is that every command in the quickstart runs in CI against the published wheel, and that the deny step asserts the error type. A deny is exactly the claim that rots without anyone noticing — a broken example still starts, still lists tools, still answers; it just stops refusing — so the gate asserts the refusal.

## What is here

- **`examples/rugpull/`** — one tool whose description comes from `RUG_DESC`, so the same server is honest on one run and poisoned on the next while nothing else about it changes. That is the version of tool poisoning that alters no parameter, and it is caught because Hangar's digest covers `description`, not only `inputSchema`.
- **`scripts/smoke_published_artifact.py`** — a new `walk_the_quickstart` step, run after the existing HTTP call and metrics assertions, driving the documented sequence over **stdio** with the SDK client:

  1. `mcp-hangar pin --write` → pins appear in the config
  2. the flat call → answers, `is_error` false
  3. restart with `RUG_DESC` changed → same call → `is_error` true and `pinned digest` in the message
  4. `mcp-hangar pin --check` → exit code 1, drift named

  The driver forwards its environment explicitly, because the SDK hands a scrubbed one to the server it spawns and `RUG_DESC` would never reach the upstream — which is a trap worth knowing about: without it the "rug pull" silently does not happen and the gate would assert a deny that never occurred.

## What is deliberately not in the gate

`mcp-hangar init`. Its first act is fetching MCP servers from npm, so a release gate that runs it goes red whenever someone else's package registry has a bad day — and a gate that fails for reasons this project cannot fix gets switched off. The configuration `init` writes is asserted in the unit suite instead (`test_init_writes_a_config_that_governs.py`, in #1198), against the real config schema. Everything from the pin onwards runs here.

## How tested

Built a wheel with `uv build --wheel` and ran the gate against it:

```
  import resolves to the artifact: 2.17.1 at …/venv/lib/python3.11/site-packages/mcp_hangar/__init__.py
  /health/live answered 200 on http://127.0.0.1:64897
  tools/call round-tripped through a cold backend (SDK v2)
  /metrics carries mcp_hangar_tool_calls_total for the call just made
  `mcp-hangar pin --write` pinned the upstream's tools
  the pinned tool answers over stdio through front_door
  the same call after a description change is refused: pinned digest mismatch
  `mcp-hangar pin --check` exits 1 and names the drifted tool

PASS — the installed artifact serves a real tool call end to end, and refuses a poisoned one.
```

`ruff check` clean on the touched files (the four pre-existing errors under `scripts/` are in `migrate_db.py` and `validate_config.py`).

### DIFF
--- a/changelog.d/1193-quickstart-deny.added.md
+++ b/changelog.d/1193-quickstart-deny.added.md
@@ -1,0 +1,13 @@
+**core:** `examples/rugpull/` is the upstream the quickstart uses to show a rug pull -- one
+tool whose description comes from `RUG_DESC`, so the same server is honest on one run and
+poisoned on the next while nothing else about it changes. A poisoned description is the
+version of tool poisoning that alters no parameter, and Hangar's digest covers
+`description` as well as the schemas, so `pin --check` catches it and the gate refuses the
+call.
+
+The release smoke gate (`scripts/smoke_published_artifact.py`) now walks that same
+sequence against the published artifact: pin, call and get an answer, restart with the
+description changed, call again and get `Tool 'echo' schema does not match its pinned
+digest`, then `pin --check` exiting 1 with the drift named. A deny is the claim that rots
+silently -- a broken example still starts, lists and answers, it just stops refusing --
+so the gate asserts the refusal rather than the happy path.
--- a/examples/rugpull/README.md
+++ b/examples/rugpull/README.md
@@ -1,0 +1,42 @@
+# The rug-pull demo upstream
+
+One tool, one environment variable. `RUG_DESC` becomes the tool's description,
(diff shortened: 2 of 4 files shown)


mapyr:

PR #1200, opened 2026-09-03
chore(core): stop showing internal identifiers in the CLI

### LINKED ISSUE #1195: WS-5: CLI hygiene -- retire the mcp_server fossils from help text
Part of #1189 (EPIC-FUNNEL). Independent of WS-1..WS-4 in content, but ships in the same release.

- `mcp-hangar --help` root description and every command help string (E8): "Production-grade MCP mcp_server platform", "Show status of all mcp_servers", "Add a mcp_server", init panel "McpServers configured".
- Grep gate: no `mcp_server` in strings rendered by Typer; snake_case stays in identifiers and config keys only. Note the known trap -- legacy aliases in this repo are string-split on purpose, so a naive grep for the canonical spelling under-reports; search the `Provider` spellings too.
- `scripts/` and `examples/` in the sdist: decide once -- keep `examples/` (WS-3 needs it), drop `scripts/` from the sdist unless something at runtime imports it (nothing found so far; confirm before dropping).

This is the one part of 2.18.0 that changes behavior for everyone (help text), so it belongs in the release note as such.

mapyr:
Closes #1195. WS-5 of #1189. **Stacked on #1199 → #1198 → #1197**; merge in that order.

## Why

```
$ mcp-hangar --help
MCP Hangar - Production-grade MCP mcp_server platform
  status      Show status of all mcp_servers.
  add         Add a mcp_server from the MCP Registry.
$ mcp-hangar status --help
Usage: mcp-hangar status [OPTIONS] [mcp_server]
```

`mcp_server` is an identifier the `provider` rename left behind. It is in the root description, in nine command help texts, in the option help of half of them, and in `status`'s `Usage:` line — every one of them read by someone who has no reason to know the codebase.

Now:

```
MCP Hangar - the policy enforcement plane for your MCP servers
  status      Show status of all MCP servers.
  add         Add an MCP server from the MCP Registry.
Usage: mcp-hangar status [OPTIONS] [SERVER]
```

The root line is the change worth arguing about: "Production-grade MCP server platform" names a category nobody searches for, in the vocabulary of the code. The replacement is the register the README and the site already use.

## Scope of the rename

Only what a renderer shows: help texts, the docstrings Typer turns into help, and the `status` metavar. Untouched: every identifier, the `mcp_servers` configuration section, and the strings that are names rather than prose.

`init --mcp_servers` becomes `init --servers`, **with the old spelling kept as an alias**. Renaming a flag people have in scripts is a breaking change; renaming the one they read is not.

## The gate

`tests/unit/cli/test_the_cli_speaks_english_not_identifiers.py` invokes the real Typer app for the root and all nine command groups and asserts on what comes back. Grepping the source would not do: a help string can be built from a constant, a docstring or a decorator argument, and only the rendered text is what someone sees. Rich wraps at the terminal width, so the check collapses whitespace before matching — otherwise a phrase split across two lines would slip through.

Two more assertions pin decisions rather than strings: the root help says "policy enforcement plane" and not "Production-grade", and `--mcp_servers` still parses.

## sdist

#1195 asked for a decision. `examples/` stays — the quickstart points at `examples/rugpull/`. **`scripts/` also stays**, contrary to the issue's guess, because it is not only a build-time thing: `tests/unit/test_api_route_inventory.py` imports `scripts.dump_api_routes`, so dropping it would ship a `tests/` directory that cannot run. Dropping both is a coherent alternative and a separate decision.

What is dropped is what has no business being in a source distribution at all: `.grimp_cache`, `.import_linter_cache`, `.clusterfuzzlite` and `coverage.json` — artefacts of having built the project here. 2.6 MB → 2.3 MB.

## How tested

`7019 passed` (unit) + `267 passed` (integration), ruff + format clean, import contracts `1 kept, 0 broken`. `uv build --sdist` inspected before and after for what the archive contains.

### DIFF
--- a/changelog.d/1195-cli-hygiene.changed.md
+++ b/changelog.d/1195-cli-hygiene.changed.md
@@ -1,0 +1,19 @@
+**core:** the CLI stopped saying `mcp_server` at people. `mcp-hangar --help` described "a
+Production-grade MCP mcp_server platform"; `status` offered to "Show status of all
+mcp_servers"; `add` would "Add a mcp_server". That spelling is an internal identifier left
+over from the `provider` rename, and it appeared in the root description, in nine command
+help texts and in `status`'s `Usage:` line. All of it now reads "MCP server(s)", and the
+root line says what Hangar is: *the policy enforcement plane for your MCP servers*.
+
+`init --mcp_servers` is now `init --servers`, with the old spelling kept as an alias --
+renaming a flag people have in scripts is a breaking change; renaming the one they read is
+not. `mcp_servers` stays where it is a name rather than prose: the configuration section
+and every identifier in the code.
+
+A test walks the real Typer app and reads what each command renders, so a help string that
+reintroduces the identifier fails the build rather than shipping.
+
+**core:** the sdist no longer carries `.grimp_cache`, `.import_linter_cache`,
+`.clusterfuzzlite` or `coverage.json` -- artefacts of having built the project, not inputs
+for building it (2.6 MB to 2.3 MB). `scripts/` and `examples/` stay: the test suite imports
+`scripts.dump_api_routes`, and the quickstart points at `examples/rugpull/`.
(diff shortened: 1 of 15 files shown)


mapyr:

PR #1205, opened 2026-09-03
fix(core): document arguments that exist

mapyr:
Follow-up to #1195, found while checking that every branch from the funnel work actually landed.

## Why

The CLI vocabulary rename rewrote `mcp_server` → "MCP server" in every string the package renders. Two of those strings were not prose:

```python
def _collect_config(mcp_server: McpServerDefinition) -> dict | None:
    """Collect configuration for an MCP server.

    Args:
        MCP server: MCP server definition      # ← documents an argument no function has
```

Same in `ConfigFileManager.add_mcp_server`. Nothing failed, because nothing parses those at runtime — which is the argument for a gate rather than against one: this is the class of edit a renamer makes silently and a reviewer skims past.

Both are corrected to the parameter's real name, and a test now checks that every `Args:` entry in the CLI package names a parameter of the function it documents. It is deliberately narrow — Google-style sections only, types and defaults ignored, continuation lines skipped — because a docstring gate that argues about style gets deleted.

## How tested

`7046 passed` (unit); ruff and format clean.

Probed rather than trusted: reintroducing the exact defect turns the gate red and names the function (`_collect_config documents ['MCP server']; its parameters are ['mcp_server']`); reverting turns it green.

## While I was there

Every branch from the 2.18.0 funnel work was checked against `main` file by file. All of it landed — including the three whose PRs (#1198, #1199, #1200) show MERGED with no content, recovered through #1201. The stale branches can be deleted; the only differences left between them and `main` are the later commits `main` has and they do not.

### DIFF
--- a/changelog.d/1195-docstring-arg-names.fixed.md
+++ b/changelog.d/1195-docstring-arg-names.fixed.md
@@ -1,0 +1,6 @@
+**core:** two docstrings documented an argument no function has. The CLI vocabulary rename
+rewrote `mcp_server` to "MCP server" in every string the package renders, and an `Args:`
+entry is not prose -- it is the parameter's name -- so `add._collect_config` and
+`ConfigFileManager.add_mcp_server` ended up describing `MCP server:`. Nothing failed,
+because nothing parses those at runtime, which is why a gate now checks that every
+documented argument is a parameter of the function it documents.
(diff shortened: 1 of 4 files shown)


mapyr:
Second commit: the same rename also ate two identifiers inside code spans — `McpServer` (the class) and `mcp_servers` (the config section) became `MCP server` and `MCP servers` in `smoke_test.build_mcp_server`'s docstring. A reader following either finds nothing: one names a class that does not exist, the other a configuration section nobody can write. The epic's own rule said snake_case stays in identifiers and config keys; this is where it did not.

Gated by what the thing is rather than where it appears: a code span in the CLI package may not hold a phrase with a space in it. That cannot fire on `McpServer`, `mcp_servers` or `mcp-hangar`, and does fire on every spelling the renamer produced. Probed both ways.

`7047 passed`.

Issue #1365, opened 2026-09-12
epic: the front door's tool list follows the fleet, not the connection


mapyr:
## Background

Diagnosed on the hosted instance, 2026-09-10: the front door's tool list is bound once, when a client connects. Warming `github` mid-session left Hangar holding 46 tools and 174 KB of catalogue while the connected client still saw zero. The client was not wrong to cache — it was told `tools.listChanged: false`.

Verified against the 2.19.0 wheel from PyPI (pins `mcp==2.0.0`, `mcp-types==2.0.0`):

- `subscriptions/listen` exists (#1027, front_door only) and already publishes the three `*/list_changed` events on its stream.
- `get_capabilities` derives all four modern flags — `resources.subscribe` and the three `listChanged` — from **one** fact: whether `subscriptions/listen` is served. There is no way to advertise the nudge without advertising the subscription.
- On the legacy handshake Hangar advertises `tools.listChanged: false` and sends no notification afterwards. That is the path Claude Code and Cursor take.
- 2.19.0 shipped #1231 as a deliberate **mitigation**, not a fix: a bounded 5 s wait on a first empty listing, only when the caller has an identity and the projection is empty because nothing has been discovered yet. Its own docstring names the complete fix — advertising `listChanged` and pushing `notifications/tools/list_changed` when a catalogue lands — and says it is a larger change.

## Why

This is the correctness half of the front-door problem and it stands alone. It changes no configuration, breaks no deployment, and needs no decision about what the surface should look like. A client that connects during warm-up, or stays connected while the fleet changes, should end up with a true list.

## Scope rule

Nothing in this epic changes the shape or size of the projected surface. That is a separate question with its own epic and its own prerequisite ADR; the two must not take each other hostage.

## Child issues

Listed in the children below.

## Acceptance criteria

- [ ] A client connected before an upstream is warm ends up with that upstream's tools without reconnecting.
- [ ] A name that leaves the projection fails in a way the client can act on.
- [ ] The change is measurable: surface size and composition are observable per replica.

## Out of scope

Opt-in flat projection, `hangar_find`, the default front-door surface, and any migration that refuses an existing config.

## Known constraint

`subscription_relay`'s own docstring records that sessions are per-replica (#877), so a subscription dies with its pod. Anything built on that path inherits it — same family as #1315 and #1358.

mapyr:
[...]

Issue #1231, opened 2026-09-09
front_door over stdio serves an empty upstream catalogue to any client that lists at startup, and tells it the list will not change


mapyr:
Found while building the WS-4 recording (#1194) from the quickstart's own configuration on **2.18.1**.

A client that connects, calls `tools/list` once, and caches the result — which is the normal thing, and the *correct* thing given what the gateway advertises — gets **only the `hangar_*` management tools**. None of the upstream's tools are in it. The upstream's tools appear about a second later, but nothing ever tells the client that.

## Reproduce

The quickstart's config, unmodified (`tool_access.mode: front_door`, one subprocess upstream, `auth.stdio.principal` with `roles: [viewer]`), then a plain SDK client over stdio:

```python
async with ClientSession(read, write, message_handler=handler) as s:
    init = await s.initialize()
    print("advertised tools capability:", init.capabilities.tools)
    first = [t.name for t in (await s.list_tools()).tools]
    await anyio.sleep(6)
    later = [t.name for t in (await s.list_tools()).tools]
```

Output:

```
advertised tools capability: list_changed=False
first list: 9 tools, echo=False
after 6s:   10 tools, echo=True
notifications received meanwhile: NONE
```

Reproducible on every run; it is not a slow first start. The first list is answered before the upstream's catalogue exists.

## Why it is more than a race

The three parts compound, and the third is what makes it a dead end rather than a hiccup:

1. **The stdio server accepts requests before the upstream is warm.** The boot log launches the subprocess and starts the stdio server in the same millisecond:

```
10:56:00.193108 Launching subprocess: [...rugpull/server.py]
10:56:00.193273 starting_stdio_server
```

2. **`initialize` advertises `list_changed=False`** on that connection, so a well-behaved client is *entitled* to list once and never poll again.
3. **No notification is ever sent** — `notifications received meanwhile: NONE` above, across six seconds in which the catalogue demonstrably grew from 9 tools to 10.

So the client is told the list is static, is handed a list that is wrong, and is never corrected. Its only recovery is reconnecting, and it has no reason to think it should.

Note the two surfaces disagree: `fastmcp_server/server_discover.py:129` answers `{"tools": {"listChanged": true}}` on the modern `server/discover` surface, while the legacy `initialize` path advertises `false` — the value observed above.

## Why this is worth prioritising

[...]

Issue #887, opened 2026-08-11
front_door: an empty projection is served with no log and no metric


mapyr:
Split from the `front_door` cold-projection parent.

When the flat projection resolves to nothing, we return `{"tools": []}` and log
nothing. The two cases that produce it are very different and indistinguishable
from the outside:

- the tenant genuinely has no tools (policy says so — a correct answer);
- the replica has discovered nothing yet (a wrong answer that will fix itself,
  or not).

An operator watching a front door that has just been rolled sees healthy pods,
a 200 response, and tenants reporting that everything vanished.

**Fix:** log at warn when a projection is empty *and* at least one configured
mcp_server is not `ready`, naming the servers. A counter split by
tenant/reason would be better still, and would make the sibling issues
measurable rather than anecdotal.

Related: core#862 asks for the same kind of visibility on the deny-all path.
Same shape of problem — a fail-closed answer that is correct-looking and
unexplained.

mapyr:
## Independent verification — 2026-08-11

Confirmed against `main` @ `c4315a1`. `_flat_list_tools` (`fastmcp_server/flat_tool_projection.py:280-304`) builds the map, converts it, and returns — there is no branch on an empty result and no log statement of any kind in the handler. The only logging anywhere in the projection path is the collision warning in `_build_flat_map` (l. 209), which by definition does not fire when there is nothing to collide.

So the two indistinguishable cases the issue names are indistinguishable in the code as well, and there is currently no metric to split them by either.

Worth adding to the acceptance criteria while this is being written: the readiness of a server is available right there — the same `registry.all()` walk that produces the empty answer can be compared against the configured fleet, so the warn line can name the servers without a second lookup. And the `cold`/`initializing` states the warning would report on are the exact ones the health worker skips (`gc.py:118`), which is what makes the empty projection persist rather than resolve itself.

Related as the issue says: #862 is the same shape on the deny-all path. Both are one-line-ish additions in different files; doing them together would give an operator a single answer to *"why is this front door serving nothing?"* — today there is no answer at all.

PR #1381, opened 2026-09-13
feat(core): make flat projection generation explicit and inspectable

### LINKED ISSUE #1367: feat(core): make flat projection generation explicit and inspectable
## Background

Part of #1365.

Today the flat projection is produced as a side effect of serving a listing, and nothing can ask what it currently holds without being a client. That is why the 2026-09-10 diagnosis needed a live probe from the outside to establish that Hangar held 46 tools while the connected client saw zero: there was no way to ask the gateway what it would project right now.

## What changes

Make projection generation an explicit, callable step whose result can be inspected independently of a `tools/list` request.

This is the foundation for the rest of #1365 — the notification work needs a defensible answer to "did the projection actually change?", and the metric work needs something to measure that is not a request.

## Acceptance criteria

- [ ] Generating the projection for a given caller identity is a separate operation from answering `tools/list`, and answering `tools/list` uses it.
- [ ] A test asserts that a projection generated after a warm-up differs from one generated before it, for the same identity — the exact condition the live diagnosis found and no test covered.
- [ ] Generation is side-effect free with respect to policy: producing a projection for inspection does not register, warm, or start anything.
- [ ] No behaviour change on the serving path; existing front-door tests pass unmodified.

## Out of scope

Exposing this over any client-facing surface. What the projection contains.

## References

mapyr:
## Why

Closes #1367. Part of #1365.

Until now the flat projection was only a side effect of serving `tools/list`. Nothing could ask what the gateway would project right now without being a client, which is why the 2026-09-10 diagnosis needed a live probe from outside. This PR is the foundation for the rest of #1365. The `tools/list_changed` work needs a defensible answer to "did the projection change?", and the metric work needs something to measure that is not a request.

## What

- **`generate_projection(tenant_id) -> Projection`** in `src/mcp_hangar/fastmcp_server/flat_tool_projection.py`. It is synchronous and needs no request, session or context.
- **`Projection`** is a frozen dataclass. It holds the `tenant_id` it was generated for, `routes` (flat name -> `(mcp_server, tool)`) and `tools` (flat name -> the definition that is served, in served order), both as read-only mappings.
  - Equality answers "did it change?". Two projections are equal when they have the same identity, the same routes and the same definitions.
  - Order is not part of equality. A re-discovery that only reorders the registry is not a change; a new definition under the same name is.
  - It is unhashable, like the mappings it holds.
- **`tools/list` uses it.** `_list_projected_tools` calls `generate_projection`, and regenerates after the #1231 warm-up wait when that wait fires. The served list is `list(projection.tools.values())`, the same objects in the same order as before.
- **One source of truth for list and call (ADR-022).**
  - The per-POST memo (`_MEMO_ATTR`) now holds the `Projection` itself, and the tenant check is `memo.tenant_id == tenant_id`.
  - On a memo hit, a flat `tools/call` routes on the routes of the projection that was just shown.
  - On a miss it builds the routes with the same `_build_flat_map` that produces `Projection.routes`. It deliberately does not build the definitions too, because that would add a schema validation of every tool to every call, which is a serving-path change.
  - `register_flat_tool_handlers` is untouched. It stays at the complexity ceiling and `ruff --select C901` passes.

### Where each side effect of the old path now lives

| Side effect | Now belongs to | Why |
|---|---|---|
| `_invalid_header_annotation`: warning and `PROJECTION_WITHDRAWALS_TOTAL` | generation | It goes through the verdict cache, so it fires once per (tool, schema digest) and never per generation. It describes the catalogue, not traffic, so whichever generation first meets a schema reports it. |
| `_denied_header_exposure`: warning and `PROJECTION_WITHDRAWALS_TOTAL` | generation | The same verdict-cache argument. |
| `_denied_header_exposure` raising `ConfigurationError` under `refuse_boot` | generation | The operator asked for the catalogue to be unavailable, not quietly smaller. An inspection that returned the smaller catalogue would misreport what a client gets. |
| Resolver `_policy_cache` fill and the throttled `front_door_denied_no_tenant` warning | generation | It memoises an answer and registers no policy. |
| `flat_tool_name_collision` warning | generation | It was logged on every listing and is logged on every generation. Unchanged. |
| `_catalogue_settled` warm-up wait (#1231) | serving | Inspection reads state and never waits. |
| `PROJECTED_TOOLS` observations and `_report_empty_projection` (#887) | serving | They measure what a client was handed. |
| Management tools, per-POST memo, cache-scope `_meta`, `PARAM_HEADER_VALIDATION_SKIPPED_TOTAL` | serving | They are per-request by nature. |

## How tested

```bash
uv venv .venv && uv pip install -e ".[dev]"      # private venv; mcp_hangar.__file__ resolves to this worktree
ruff check src tests                             # All checks passed
ruff format --check src tests                    # 962 files already formatted
ruff check src --select C901                     # All checks passed
mypy src/mcp_hangar                              # Success: no issues found in 429 source files
python scripts/check_dead_symbols.py             # baseline holds: 12 unreferenced, 25 exported-unreferenced
python scripts/build_changelog.py check          # OK: 1 changelog fragment(s) valid
pytest $(grep -rl flat_tool_projection tests/unit)   # 333 passed: every existing front-door test, unmodified, plus the new file
pytest tests/unit                                # 7616 passed, 273 skipped, 1 failed (pre-existing, see below)
MCP_HANGAR_LIVE_VERIFY=1 pytest tests/live -m "live and t0"   # 21 passed
```

New tests are in `tests/unit/test_the_projection_is_generated_apart_from_the_listing.py`. They use the real `ToolProjectionRegistry`, the real resolver, the real `catalogue_warmup` state, the real `lifecycle.warm_the_front_door_catalogue`, and the real `McpServerStarted` handler that fills the registry. The only stand-in is the upstream start itself.

- **Before vs. after warm-up.** For the same identity, a projection generated after the warm-up differs from one generated before it. This is the condition the live diagnosis found.
- **Mid-warm-up.** A projection generated while the warm-up is in flight comes back at once, in under 1 s, with what exists, and it differs from the one generated after.
- **No side effects.** Generating for a tenant, another tenant and no identity sends no command through either process-global start path (`get_command_bus`, `get_runtime`). It also leaves the catalogue unchanged, registers no policy, starts no warm-up, and leaves the cold server cold.
- **Not counted as a listing.** `PROJECTED_TOOLS` and `EMPTY_PROJECTION_TOTAL` do not move for empty projections (nothing discovered, no identity) or for a non-empty one.
- **`tools/list` serves the generated projection.** Through the registered SDK v2 handler, and also after the listing waits out a real warm-up.
- **What "changed" means.**
  - Regenerating an unchanged catalogue is no change.
  - A reorder-only re-discovery is no change.
  - A changed definition under the same routes is a change.
  - The same surface for another identity is a different projection.
- **The tests bite.** With `_report_empty_projection` temporarily moved into `generate_projection`, the not-counted test failed. The mutation was reverted before committing.

**Failures and gaps, stated plainly**

- **`test_baggage_scrub.py::TestTheChokepoint::test_a_carrier_loses_its_baggage_even_with_tracing_off[sdk]`** fails with `AttributeError: 'NoneType' object has no attribute 'get_tracer_provider'` (`observability/tracing.py:655`).
  - It fails identically on the untouched branch and when run on its own, so it is environmental in this venv (the OTel SDK extra is not installed) and unrelated to this change.
- **`test_there_are_dockerfiles_to_check`** lives in `tests/ci`, not `tests/unit`, so it was not part of these runs. It is known to fail under `.claude/worktrees` for environmental reasons.
- **Live coverage.**
  - The T0 tier ran, 21 passed. Its front-door tests (`test_t0_stdio_funnel`, `test_t0_init`, `test_t0_pin`) drive the front door over **stdio**, or check config and pins.
  - No T0 test lists per-tenant front-door tools over streamable HTTP. The T2 tier (`test_t2_auth`, front-door DENY) needs Keycloak and was **not run**.
  - So the per-request listing path is not live-verified over streamable HTTP by this PR. The serving path's behaviour is unchanged, but this verification gap is real.

## Risk and rollback

Low risk. The serving output is the same objects in the same order, the memo semantics are the same, and no config, wire or metric changes. Revert the single squash commit.

## Agent metadata

- [x] Single goal (no scope creep): nothing is exposed on a client-facing surface, and what the projection contains is unchanged
- [x] LOC declared upfront: src +123/-17, most of it docstrings; tests +316; changelog +7


mapyr:
Reviewed: `tools/list` serves exactly what `generate_projection` returns. The per-POST memo now holds the `Projection` and cannot answer for another tenant. The only other readers of the memo are tests that patch these functions, and their stand-ins still fit the new signatures. I pushed one fix, a missing full stop in the changelog fragment. CI is fully green, including `test (3.11-3.14)`, which confirms that the local `test_baggage_scrub` failure the author hit was environmental (the OpenTelemetry SDK was missing from the local venv). Marking ready.

Issue #1360, opened 2026-09-12
fix(core): record giving up on a server as a stop with its own reason


mapyr:
## Background

Split out of #1325.

## Why

`mcp_hangar_mcp_server_stops_total` carries a `reason` label, and on a live installation the only reason ever observed is `idle`. Grepping the call sites: `gc.py` passes `"idle"`, the stop command handler passes `reason or "manual"`, and the metrics event handler mirrors `event.reason`. Nothing records "stopped because health checks failed".

So the one event that matters most — the gateway giving up on a provider — is either invisible or indistinguishable from routine idle reaping, and `increase(mcp_hangar_mcp_server_stops_total{reason!="idle"}[1h]) > 0` cannot be written.

## What changes

Record the give-up as a stop with its own reason, e.g. `record_mcp_server_stop(id, "health_check_failed")`, so the give-up path emits a distinguishable counter increment.

## Acceptance criteria

- [ ] The health-check give-up path emits a stop with a dedicated reason, distinct from `idle` and `manual`.
- [ ] A test drives a provider past `max_consecutive_failures` and asserts the counter increments with that reason.
- [ ] Existing reasons keep their current values; no relabelling of `idle` or `manual`.
- [ ] The set of reasons is documented where the metric is documented, so an operator can write `reason!="idle"` rules against a closed set.

## Out of scope

The `dead` vs `cold` state enum question, and the last-healthy timestamp (#1359). Chart alert rules live in `mcp-hangar/helm-charts`.

## References

- https://github.com/mcp-hangar/mcp-hangar/issues/1325

Issue #1358, opened 2026-09-12
feat(core): share group rotation and circuit state across replicas


mapyr:
## Background

Split out of #1315. This is the design half; the three mechanical halves are #1355, #1356 and #1357.

## Why

A group's rotation and circuit-breaker state live in each replica's memory. Three consequences, all observed:

- `GET /api/groups/{id}` and `hangar_details` answer differently depending on which pod serves the request. In our environment this read as "production is broken" for several minutes while production was fine.
- A client pinned to the tripped replica gets `NoAvailableMemberError` indefinitely. No metric anywhere reports it, because the peers look healthy.
- The breaker's own purpose is diluted: with N replicas an upstream must trip each one separately, so protection engages N times later than configured.

Same family as #877 (sessions are per-replica), different symptom.

## What changes

Decide, then implement. The decision is the deliverable here:

1. Share rotation and circuit state across replicas, using the coordination this project already runs for management leases — a group is open or closed once, not per pod.
2. Or keep state per-replica and accept it as the contract, documenting that the breaker threshold is per-pod and that reads are replica-local.

Option 1 changes what `max_consecutive_failures` means (fleet-wide rather than per-pod) and puts breaker transitions on the event log, which brings the commit-ordering and projection/effect taxonomy rules into scope. Option 2 is cheap but leaves the pinned-client outage shape in place, mitigated only by #1355.

Blocked on #1355 landing first: a replica that cannot heal itself is a bug regardless of which option wins, and fixing it may change how bad option 2 looks.

## Acceptance criteria

- [ ] An ADR records the decision, the meaning of `max_consecutive_failures` under it, and what a replica-local read is allowed to report.
- [ ] If shared: a test proves two replicas agree on circuit state within a bounded window after a trip on one of them.
- [ ] If shared: breaker transitions respect the existing handler taxonomy (projection vs effect vs local view) and the commit-ordered log — no new ordering assumptions.
- [ ] If per-replica: the contract is documented where an operator reads group status, and #1357's divergence metric is referenced as the supported way to detect it.

## Out of scope

Session affinity, and `#877`.

## References

- https://github.com/mcp-hangar/mcp-hangar/issues/1315
- https://github.com/mcp-hangar/mcp-hangar/issues/877

mapyr:
**Decision (maintainer, 2026-09-13)**, recorded together with #1380.

[...]

Issue #1315, opened 2026-09-11
Group circuit breaker is per-replica and cannot close itself: one pod serves NoAvailableMemberError indefinitely while its peers are healthy


mapyr:
A group's rotation and circuit-breaker state live in each replica's memory. One
replica can hold a group permanently open while its peers serve it, and the
affected replica does not recover on its own -- not from passing health checks,
not from traffic, not from time. Only `hangar_group_rebalance` or a restart
brings it back.

## Reproduction

Environment: 2 gateway replicas, one group (`weather`, `strategy: priority`,
three remote members, `min_healthy: 1`).

1. Cut the members' egress so every call fails (I used a CiliumNetworkPolicy
   `egressDeny`; anything that makes the upstream unreachable will do).
2. Call a tool through the group ~14 times against ONE replica.
3. Restore egress.
4. Ask both replicas for the group.

Opening behaves exactly as documented -- two consecutive failures per member,
members leave rotation, the circuit opens:

```
14 calls: 6x ToolInvocationError, then 8x NoAvailableMemberError
circuit_open: True | is_available: False | state: degraded | healthy_count: 0
members: all `ready`, in_rotation False, consecutive_failures 2
```

## 1. The two replicas disagree, at the same instant

Queried directly, pod by pod, seconds after egress was restored:

```
mcp-hangar-...-949g4 -> circuit_open=True   is_available=False  healthy=0  in_rotation=[F,F,F]
mcp-hangar-...-kjpkq -> circuit_open=False  is_available=True   healthy=3  in_rotation=[T,T,T]
```

`GET /api/groups/weather` therefore answers differently depending on which pod
the request lands on, and so does `hangar_details`. In our environment this
read as "production is broken" for several minutes while production was fine --
the failures had all landed on the replica our session was pinned to.

## 2. The tripped replica never recovers

Five minutes after the upstreams were healthy again, that replica still reported
`circuit_open: True, healthy_count: 0`, with `consecutive_failures: 2` on every
member. During that window its own log contains **seven** `HealthCheckPassed`
events for those very members, and I drove six more tool calls through it (all
`NoAvailableMemberError`).

`consecutive_failures` staying at 2 is the tell: `report_success()` sets it to 0
as its first act, so it was never called on that replica -- even though
`GroupRebalanceSaga` is supposed to call it on `HealthCheckPassed`. Whatever the
delivery path is, it did not reach the group object on the replica that needed
it.

Then, on that same replica:

```
[...]

PR #1383, opened 2026-09-13
fix(core): return a member to group rotation on a passing health check

### LINKED ISSUE #1355: fix(core): a passing health check must return a member to group rotation
## Background

Split out of #1315, from a live 2-replica environment (core 2.18.2, group `weather`, `strategy: priority`, three remote members, `min_healthy: 1`).

## Why

A replica that holds members out of rotation cannot put them back by itself. Passing health checks do not do it, traffic does not do it, time does not do it. Today the only recovery is `hangar_group_rebalance` or a pod restart, and with session affinity an operator cannot choose which replica they reach — so in practice the remedy is "restart the deployment".

## What changes

Find and fix the delivery path so a `HealthCheckPassed` for a member reaches the group object on the replica that observed the failures.

The tell is in the measured state: five minutes after the upstreams were healthy again, the tripped replica still reported `circuit_open: True, healthy_count: 0`, with `consecutive_failures: 2` on every member, while its own log carried **seven** `HealthCheckPassed` events for those very members and six further tool calls returned `NoAvailableMemberError`.

`report_success()` sets `consecutive_failures` to 0 as its first act. It staying at 2 proves the method was never called on that replica, even though `GroupRebalanceSaga` is supposed to call it on `HealthCheckPassed`. `rebalance()` has exactly one caller in the codebase — the `hangar_group_rebalance` tool — and invoking it recovers the group instantly:

```
hangar_group_rebalance(weather)
  -> {"group_id": "weather", "state": "healthy", "healthy_count": 1,
      "members_in_rotation": ["weather-open-meteo"]}
circuit_open: False
```

So the recovery logic is correct and reachable; only the event path into it is missing.

## Acceptance criteria

- [ ] Integration test: drive a member out of rotation through consecutive failures, deliver `HealthCheckPassed` for it, and assert `in_rotation` returns to `True` and `consecutive_failures` is 0 — without calling `hangar_group_rebalance`.
- [ ] The same test asserts a subsequent tool call through the group succeeds rather than returning `NoAvailableMemberError`.
- [ ] A test covers the multi-member case: recovery of one member is enough to close the circuit when `min_healthy: 1`.
- [ ] The fix does not introduce a second caller of `rebalance()` on the hot path; recovery happens through `report_success()`.

## Out of scope

Sharing rotation or circuit state across replicas — that is the design question, tracked separately under #1315. This issue makes a single replica able to heal itself, which is the smaller half and is a straight bug.

## References

mapyr:
## Why

Closes #1355. A group member driven out of rotation by consecutive failures never came back on its own. Passing health checks, traffic and time all left it out; only `hangar_group_rebalance` or a restart recovered the group. The recovery logic in `report_success()` was fine. On the path `serve` runs, `GroupRebalanceSaga` never found the member's group, so every `HealthCheckPassed` stopped at its lookup.

Root cause, shown on unfixed `main` against the saga that the real `bootstrap()` registered, not one built by hand:

- `bootstrap()` runs `load_config` (`server/bootstrap/__init__.py:457`) before `init_saga` (`:510`). So `_load_group_members` (`server/config.py:215`, `:280-281`) got `None` from `get_group_rebalance_saga()` and registered no member. Observed: `saga._member_to_group == {}`.
- `init_saga` passed `ctx.groups` (`server/bootstrap/cqrs.py:337`). At that point this is still the empty dict `init_context` built (`server/context.py:276`). `ctx.groups = GROUPS` only happens at `bootstrap/__init__.py:661`. Observed: `saga._groups is GROUPS` is `False`, and `len(saga._groups)` is 0.
- So `_get_group_id(member)` returned `None`, and `handle()` returned at `group_rebalance_saga.py:113-115`. Both orderings go back to #33; this is not a regression from #807.
- A third gap sat behind those two: only `rebalance()` closed an open group circuit. `report_success()` never touches the breaker, and the group never calls `allow_request()`, so `circuit_open: true` and `degraded` stuck even after a member was back.

Joining the path up made every event the saga handles reach the group for the first time. The saga counted two things wrong, and review caught both before this PR was marked ready:

- **Every stop counted as a failure.** Idle reaping included, which could take a cold member out of rotation, where nothing selects, health-checks or starts it again.
- **One failing health check could count twice.**

## What

- `GroupRebalanceSaga` reads a member's groups from the live groups mapping on every event (`_groups_of`) and reports to each group the member is in. The `register_member` table is now consulted only when the saga has no groups mapping.
- `init_saga` hands the saga `GROUPS`, the dict the executor and the CRUD handlers already use, instead of `ctx.groups`.
- `McpServerGroup.report_success()` closes an open circuit once `min_healthy` members are in rotation (`_maybe_close_circuit`) and emits `GroupCircuitClosed`.
- A stop no longer counts as a failure. `McpServerStopped` only logs; the reasons are enumerated below.
- A failing health check counts once. `McpServerDegraded` whose reason is `DEGRADED_BY_HEALTH_CHECKS` is not counted, because that same check's `HealthCheckFailed` already was. The constant is new and names the existing `"health_check_failures"` string, which is still what the event carries, so both places use one name. Any other degrade, from a failed start, still counts once.
- There is no new caller of `rebalance()`. Recovery goes through `report_success()`.

### Every reason `McpServerStopped` carries

`McpServer._shutdown_internal` (`domain/model/mcp_server.py:1859`) is the only place the event is built, and it only ever receives these two reasons:

| reason | emitted by | callers | a failure? |
|---|---|---|---|
| `idle` | `maybe_shutdown_idle()` (`mcp_server.py:1830`) | the GC worker (`gc.py:108`) | no: the gateway reaped an unused server |
| `shutdown` | `shutdown()`/`stop()` (`mcp_server.py:1838`, `:1842`) | `StopMcpServerCommand` handler (`application/commands/handlers.py:117`), config reload (`reload_handler.py:132`), CRUD delete (`crud_handlers.py:383`), hot unload (`load_handlers.py:457`), `McpServerService.stop` (`mcp_server_service.py:110`), group `stop_all` (`mcp_server_group.py:759`) | no: an operator or the gateway chose it |

No stop reason means "the server failed".

- The `StopMcpServerCommand` handler writes `command.reason` (`"manual"` by default, or the recovery saga's `"max_retries_exceeded"`) only to a metric. The event still says `"shutdown"`.
- The give-up stop always follows an `McpServerDegraded`, which has already counted. Giving it a distinct reason is #1360.
- A crashed process emits no stop at all. `health_check` marks it `DEAD` without an event (`mcp_server.py:1741`).

Failures still reach the group in three ways: `HealthCheckFailed`, `McpServerDegraded` from a failed start, and the executor's `report_failure` on a failed call.

### Double counting

When a check reaches the server's degrade threshold, `health_check()` emits `HealthCheckFailed` and then `McpServerDegraded(reason="health_check_failures")` for that one check (`mcp_server.py:1771-1797`). Both used to reach `report_failure`.

A start that fails during a group call does not double-count. `_gate_cold_start` (`executor.py:1477`) refuses the call before the executor's report block runs, so the degrade from `_handle_start_failure` (`mcp_server.py:1262`) is the group's only signal for it.

## How tested

I wrote each test before its fix, against the served wiring. `tests/integration/test_a_passing_health_check_returns_a_member_to_rotation.py` runs `_group_recovery_harness.py` in a fresh interpreter. The harness uses:

- the real `bootstrap(config_dict=...)`;
- `hangar_call` and `hangar_group_list` through `mcp_app_for_serving`, the app `serve --http` serves;
- `tests/mock_provider.py` members over stdio;
- the background workers `bootstrap()` created.

Nothing in the harness builds a saga, registers a member, or publishes an event by hand. The worker intervals are lowered before bootstrap reads them. `McpServerGroup.rebalance` is wrapped only to count its calls, and the test asserts the count is 0.

There are three modes:

- `single`: a one-member group, driven out by failing calls and then left to the health-check worker.
- `pair`: the same with two members and `min_healthy: 1`, where only one member recovers.
- `idle`: two members with `idle_ttl_s: 1` and the strictest thresholds a group takes (`unhealthy_threshold: 1`, circuit `failure_threshold: 2`). The real GC worker reaps them and group calls start them again, three times over.

Results:

- **Recovery, on unfixed `main`:** 6 failed, 6 passed. The failures reproduce the live report: three `HealthCheckPassed` events for the member, yet `in_rotation: false`, `consecutive_failures: 2`, `circuit_open: true`, and the next call returns `NoAvailableMemberError`.
- **Recovery, each src change reverted on its own:**
  - Reverting the `cqrs.py` change: 6 failed.
  - Reverting the saga lookup change: 6 failed.
  - Reverting the group change: only the two circuit tests fail.
- **Idle, on the saga as first pushed:** the four idle tests fail. After the first reap both members are out of rotation, the circuit is open, and the cycle 2 calls are refused.
- **Unit tests,** `tests/unit/test_a_passing_health_check_reaches_the_group.py`, 11 tests:
  - A stop with either reason leaves the member in rotation and the circuit closed.
  - One real failing `McpServer.health_check()` counts once, even when it also degrades the server.
  - A failed start that degrades the server counts once.
  - Membership is read from the live groups mapping.
  - The circuit closes once `min_healthy` members are back, and not below.

  On the saga as first pushed, 3 of the 11 fail, including `consecutive_failures == 4` where 3 are expected.
- **Fixed:** all 28 new tests pass (17 integration, 11 unit).

I ran the gates below locally, in a private venv inside this worktree, installed the way CI installs (`.[dev,opentelemetry]`). `python -c "import mcp_hangar; print(mcp_hangar.__file__)"` printed `.../.claude/worktrees/agent-a808864106a2c7d4f/src/mcp_hangar/__init__.py`. I captured each exit code directly rather than through `| tail`.

```text
ruff check src tests                       # exit 0
ruff format --check src tests              # exit 0 (964 files already formatted)
mypy src/mcp_hangar                        # exit 0 (no issues found in 429 source files)
pytest tests/integration/ --timeout=60     # exit 0: 397 passed, 5 skipped
pytest --ignore=tests/integration          # exit 1: 8002 passed, 63 skipped, 1 failed
make changelog-check                       # exit 0
```

The one unit failure is `tests/ci/test_base_images_are_pinned_and_updated.py::test_there_are_dockerfiles_to_check`, which always fails under `.claude/worktrees` because of the environment.

## Risk and rollback

- **Events that now count as member failures:** `HealthCheckFailed`, one per failing check; and `McpServerDegraded` from a failed start. These come on top of the executor's existing `report_failure` on a failed call through the group. A failing health check can now take a member out of rotation and counts toward the group circuit. Before this PR none of this happened on the served path, because the saga never found a group.
- **Events that do not count:**
  - `McpServerStopped`, whatever its reason: idle reaping, `hangar_stop`, reload, unload, delete, `stop_all`. A stopped member stays in rotation and is started again by the next call through the group.
  - `McpServerDegraded` caused by health checks, which its `HealthCheckFailed` has already counted.
- **Circuits close on recovery.** An open group circuit now closes on a success once `min_healthy` members are in rotation, instead of waiting for `hangar_group_rebalance`. Member selection already ignored an open circuit while any member was in rotation, so routing is unchanged. What changes is `circuit_open`, `state` and `is_available` in `hangar_group_list`, plus a `GroupCircuitClosed` event.
- **Rollback:** revert the commits.

Out of scope, and left alone on purpose:

- `_restore_group_circuit_breakers(saga_state_store, ctx.groups)` (`cqrs.py:353`) still receives the empty dict, so persisted group circuit state is never restored on the served path. It also loads by `saga_type` alone, so if it were revived as it stands, every group would get the same row.
- While closed, the group circuit's failure count never resets on success. So counted failures, now including health-check failures, accumulate over the process lifetime until the circuit opens. The next success in rotation now closes it again.
- A member whose server degrades while still in rotation stays in rotation until its failures reach `unhealthy_threshold`. Health checks stop at `DEGRADED`, and the executor's per-member gate (`executor.py:1376`) refuses without reporting to the group. This only arises when `unhealthy_threshold` exceeds the server's degrade threshold of 3, and it was already true before this PR.
- A distinct give-up stop reason (#1360), and sharing rotation or circuit state across replicas (#1315, #1358).

## Agent metadata

- [x] LOC declared upfront: see `git diff --shortstat origin/main`; src under 120 lines changed


mapyr:
Blocker, so do not mark this ready yet. This PR makes the saga live on the served path, and the saga reports every `McpServerStopped` as a member failure, whatever the `reason`. The GC worker idle-reaps group members too (`gc.py:108-111`), so every idle reap counts against the group's circuit breaker, whose failure count never resets while it is closed. Once the circuit opens, group calls are refused, cold members produce no `HealthCheckPassed`, and `_maybe_close_circuit()` never runs. The group wedges: the #1355 symptom again, reached through idle instead of failures. The fix (benign stops are not failures, plus a check for double counting and a served-path idle-reap test) is in progress on this branch.

mapyr:
The blocker is resolved in b8232543. I've reviewed both commits and am marking this ready.

- **A stop only logs now.** `McpServerStopped` carries only `idle` or `shutdown`, and the gateway or an operator chose both. A crashed process emits no stop.
- **A health-check degrade is not counted a second time.** The failing check that caused it was already reported as `HealthCheckFailed`. The shared `DEGRADED_BY_HEALTH_CHECKS` constant keeps the emit site and the counter in step.
- **Any other degrade counts once.** A failed start, for example, is the group's only signal.
- **The wedge is pinned by a served-path test.** It drives the real GC worker: members are idle-reaped and restarted three times, stay in rotation with 0 failures, and the circuit stays closed. That test failed on the first commit.
- **The recovery tests still pass.** CI is fully green.

Out-of-scope findings from this work will be filed as separate issues.

PR #1385, opened 2026-09-13
fix(cli): make pin honour --quiet, MCP_LOG_LEVEL and logging.level

### LINKED ISSUE #1236: mcp-hangar pin ignores --quiet, MCP_LOG_LEVEL and logging.level, so its output arrives buried in debug chatter
Found while recording the WS-4 demo (#1194) on **2.18.1**.

`mcp-hangar pin` writes `[info]`/`[debug]` stdio-client lines around the two or three lines that are its actual output, and **none of the three documented ways to turn that down has any effect**.

## Reproduce

```console
$ mcp-hangar --quiet pin --config demo.yaml --write
2026-09-09 13:03:54 [info     ] stdio_client_process_exited    exit_code=-15 expected=True
pinned demo: 1 tool(s)
previous config kept at demo.yaml.bak; comments and key order are not preserved

$ MCP_LOG_LEVEL=ERROR mcp-hangar pin --config demo.yaml --check
2026-09-09 13:04:12 [info     ] stdio_client_process_exited    exit_code=-15 expected=True
drift demo.echo
  pinned:   2970199253016cbcebf2d4b43d194f2cdd4c28a8517381ae8a81aecc5edff245
  serving:  9bf25830fe16061e675517741fade793af60580c30066d53b9723c632b61bca1
```

`demo.yaml` carries `logging: {level: ERROR}` in both runs -- and that same key *is* honoured on the `serve` path with the same file, so all three controls are being ignored specifically here.

A full `--check` run prints about a dozen lines before `drift`:

```
[info ] stdio_client_reader_started      pid=16015
[info ] stdio_client_sending_request     alive=True method=initialize pid=16015
[debug] stdio_client_request_sent
[debug] upstream_legacy_era              mcp_server_id=demo negotiated_protocol_version=2025-11-25
[debug] stdio_client_notification_sent   method=notifications/initialized
[info ] stdio_client_sending_request     alive=True method=tools/list pid=16015
[debug] stdio_client_request_sent
[info ] mcp_server_started: demo, mode=subprocess, tools=1
[debug] stdio_client_shutdown_rpc_failed error=client_closed
[debug] stdio_client_eof_on_stdout       expected=True
[info ] stdio_client_process_exited      exit_code=-15 expected=True
drift demo.echo
```

## Why this is more than cosmetics

`pin --check` is the command this project tells people to automate. `getting-started/quickstart.md` ships it as a pre-commit hook:

```yaml
      - id: mcp-hangar-pin
        entry: mcp-hangar pin --check
```

A hook whose real signal is `drift <server>.<tool>` plus exit 1, wrapped in a dozen debug lines that no flag removes, is a hook people will silence wholesale -- and silencing stderr loses the drift message too, because the result and the noise are not cleanly split across streams.

It also means `--quiet` does not do what `--help` says it does ("Suppress non-essential output"). An `[info]` line reporting that a subprocess exited *as expected* is the definition of non-essential.

## Suggested shape

- Have `pin` honour the logging configuration the `serve` path already honours, and make `--quiet` mean it.
- Put `pin`'s result on **stdout** and everything else on stderr, so `mcp-hangar pin --check >report 2>/dev/null` is a reasonable thing to write in CI.

Adjacent: this is why the WS-4 recording pins off camera -- a twenty-second asset cannot spend six seconds on log lines.

mapyr:
## Why

Closes #1236.

`mcp-hangar pin` never called `setup_logging`. Everything it logged went through the import-time structlog config, which has no level filter. So every stdio-client `[info]` and `[debug]` line printed around the command's two or three lines of output. None of the documented controls changed that: `--quiet`, `MCP_LOG_LEVEL`, or the config file's `logging.level`. Yet `serve` honours `logging.level` from the same file. The quickstart ships `pin --check` as a pre-commit hook, so the noise buried the one signal that matters there, the drift report.

Root cause, on `main`:

- `src/mcp_hangar/server/cli/commands/pin.py:41-100`: nothing on the path sets up logging. `serve` does it in `lifecycle.py`, where `run_server` calls `_setup_logging_from_config`.
- `src/mcp_hangar/logging_config.py:48-51`: the pre-setup config, which `pin` was running on. It is `PrintLoggerFactory(file=sys.stderr)` with structlog's default wrapper, which filters nothing.
- `src/mcp_hangar/gc.py:25`: a DEBUG line logged at import. Every CLI invocation imports gc.py through `mcp_hangar/server/__init__.py` and bootstrap, before any command body runs. No control could reach that line, even with the fix above.

The streams were already mostly split: results on stdout, logs on stderr. The only thing on stdout that did not belong there was the usage errors that end in exit 2.

### Precedence

I pulled `resolve_logging_settings` out of lifecycle's `_setup_logging_from_config`. `serve` and `pin` both call it now, so they cannot drift apart again. The order, highest first:

1. `--log-level`, new on `pin` with the same option spec as on `serve`;
2. `MCP_LOG_LEVEL`;
3. the config file's `logging.level` (and `logging.file` and `logging.json_format`);
4. INFO.

**Docstring correction.** The old `_setup_logging_from_config` docstring gave the order as CLI, then config file, then environment, then defaults. The code never worked that way. Typer, and the legacy argparse path, fold `MCP_LOG_LEVEL` into the flag's value, so the environment variable outranks the config file. The docstring now gives the real order. It also names an existing quirk: a flag or env value of exactly `INFO` looks the same as the default, so `logging.level` overrides it. `serve` behaves exactly as before.

### `--quiet`

It works both as the global flag before the subcommand (`mcp-hangar --quiet pin ...`, the issue's repro) and as a new local `-q/--quiet` after it. The local form lets a pre-commit `args: [--quiet]` work.

- **Drops:**
  - All log output below ERROR, whatever level the other controls chose. `--quiet` sets a floor rather than a level.
  - The lines that only confirm success: `every pinned tool still matches its pin.`, `pinned <server>: N tool(s)`, and the `.bak` backup note.
- **Keeps:**
  - The digests from a plain run.
  - The drift report.
  - The unreachable-server lines.
  - `--json` output, which is never trimmed.
  - Log errors, which still reach stderr.

### Streams

The answer goes to stdout: digests, drift, what `--write` did, and unreachable servers. Log output and usage errors go to stderr. The usage errors are:

- `--write` together with `--check`;
- a missing file;
- invalid YAML;
- a document that is not a mapping;
- an unknown `--server`;
- no servers configured.

`mcp-hangar pin --check >report 2>/dev/null` now keeps the drift report. Exit codes are unchanged: 0, 1 and 2 mean what they meant before.

### The watchdog notice

gc.py's "watchdog package not installed" notice is no longer logged at import. It now fires in `ConfigReloadWorker.start()`, still at DEBUG. It fires only when hot reload asked for watchdog and has to fall back to polling, so it still tells a `serve` operator why reload polls. It no longer prints on `pin` or `--help`.

## How tested

```text
ruff check src/mcp_hangar <new test files>           exit 0
ruff format --check src/mcp_hangar <new test files>  exit 0
mypy src/mcp_hangar                                  exit 0 (no issues in 429 source files)
pytest --ignore=tests/integration                    exit 1: 8018 passed, 63 skipped, 1 failed
pytest tests/integration/test_pin_output_is_quiet_and_split.py \
       tests/integration/test_pin_writes_and_detects_drift.py \
       tests/integration/test_config_reload_worker.py            exit 0: 14 passed
make changelog-check                                 OK: 1 changelog fragment(s) valid
```

The one unit failure is environmental: `tests/ci/test_base_images_are_pinned_and_updated.py::test_there_are_dockerfiles_to_check` fails with `assert 0 >= 4` because the glob finds no Dockerfiles under `.claude/worktrees`. It fails the same way on any branch run from a worktree.

My first full run was in a venv without the `opentelemetry` extra. It also failed `tests/unit/test_baggage_scrub.py::TestTheChokepoint::test_a_carrier_loses_its_baggage_even_with_tracing_off[sdk]`, at `tracing.py:655`, which needs the SDK. CI installs `.[dev,opentelemetry]`, and with that extra installed the test passes. The same first run included the full integration suite: 342 passed, 47 skipped, exit 0. That was before the gc.py notice moved.

### New tests

`tests/unit/cli/test_pin_honours_the_log_controls.py` runs in-process. The server is stubbed, and the stub logs through the real `mcp_hangar.stdio_client` logger name. It covers:

- **Baseline:** the client lines appear when no control is set, so the absence checks below cannot pass trivially.
- **Each control silences the client info and debug lines:** global `--quiet`, `pin --quiet`, `MCP_LOG_LEVEL=ERROR` and `logging.level: ERROR`. Log errors still show.
- **The issue's repro:** `--quiet pin --config X --write` with global `--quiet`, global `-q` and `pin -q`.
- **`--quiet` is a floor.**
- **Precedence** (flag, then env, then file, then default), run through the CLI and checked directly on `resolve_logging_settings`.
- **Streams:** drift on stdout and not on stderr; a usage error on stderr.
- **What `--quiet` keeps** (drift, digests, `--json`) **and drops** (the clean-check line, the write summary).
- **Import noise:** importing the CLI logs nothing, checked in a fresh subprocess.
- **The watchdog notice:** it fires from `ConfigReloadWorker.start()` when watchdog was requested, and not otherwise.

`tests/integration/test_pin_output_is_quiet_and_split.py` runs the real CLI entry point as a subprocess against `tests/mock_provider.py`:

- The issue's `--quiet pin --write` leaves both streams empty.
- For each of the three controls, a drifted `--check` exits 1, stdout starts with `drift math.add`, and stderr is empty.
- A no-control baseline shows `stdio_client_` lines on stderr and the drift only on stdout.

Before the gc.py and global-flag additions, I ran the new tests against `origin/main`'s `pin.py` and `gc.py`: 17 of 26 failed. The failures cover every control test, the precedence, quiet and usage-error tests, and the import-noise test. The 9 that passed test behaviour that was already correct: drift on stdout, `--json`, digests, the resolver table, and the integration baseline.

### Manual check

I ran the real CLI against the mock provider with `logging.level: ERROR` in the config:

| Run | Before | After |
| --- | --- | --- |
| plain | exit 0, 27 log lines on stderr | exit 0, stderr empty |
| `--write` | exit 0, 21 log lines on stderr | exit 0, stderr empty |
| `--check` | exit 0, 27 log lines on stderr | exit 0, stderr empty |
| drifted `--check` | exit 1, 27 log lines on stderr | exit 1, stderr empty |

## Risk and rollback

Low. `serve` resolves its log level with the same code, moved into a function, and behaves as before. One visible change: usage errors now go to stderr instead of stdout, which matters only to someone reading them from stdout. To roll back, revert the single squash commit.

## Docs

The quickstart's pre-commit hook (`entry: mcp-hangar pin --check`, in the separate docs repo) keeps working unchanged. It could now suggest `args: [--quiet]` so a failing hook shows only the drift report. That change is optional and not made here.

## Out of scope, noticed

- `pin` ignores the global `--config` (`mcp-hangar --config X pin`). It reads only its own `--config` or `$MCP_CONFIG`, whereas `serve` honours both.
- `pin` does not read the global `--verbose` or `--json`, and `serve` does not read the global `--quiet`.
- `pin` has no `--json-logs` or `--log-file` flag and does not read `MCP_JSON_LOGS`. The config file's `logging.json_format` and `logging.file` are honoured through the shared resolver.
- The resolver's config preload does not catch `ConfigurationError`, which an unset `${VAR}` in the file raises. `serve` has the same gap, and `pin` exits 2 either way.
- `infrastructure/discovery/filesystem_source.py` logs the same kind of watchdog notice at import. It is not on the CLI's import path.
- `ConfigReloadWorker._start_watchdog` has a `watchdog_not_available_falling_back_to_polling` warning that can never run, because `use_watchdog` already accounts for whether watchdog is available.
- `test_baggage_scrub.py::...[sdk]` has no `otel_sdk` marker, so without the extra it fails instead of skipping.
- With the `opentelemetry` extra installed, a full unit run prints a few `scheduled_command_failed ... No handler registered for StartMcpServerCommand` lines past pytest's capture. They come from saga timer threads (`infrastructure/saga_manager.py:128`) that fire after their tests have finished. They are not caused by this PR's tests: I counted 7 with the new unit file and 8 without it. `test_sagas.py` and `test_capability_drift.py` print none when run on their own. None showed up in the run without the extra. I have not compared against `main`.

## Agent metadata

- [x] Single Conventional Commit scope: `cli`
- [ ] LOC declared upfront: src +144/-35, tests +400, changelog +20. Counting the tests, that is over the ~400-line agent-friendly guide.


mapyr:
Reviewed and ready.

- `serve` and `pin` now resolve logging through one function, so one command cannot honour a control the other ignores.
- `--quiet` is a floor of ERROR, and the global flag is read through `ctx.obj`.
- The watchdog notice moved to `ConfigReloadWorker.start()`, so it fires only when hot reload asked for watchdog; it was not deleted.
- `logging.getLevelNamesMapping()` needs Python 3.11, which is the floor of the CI matrix.

I removed `agent-friendly`: with its tests, the PR is past that label's 400-line limit. CI is fully green.

Issue #1378, opened 2026-09-12
fix(core): hangar_status renders a group as [?] and its frame does not close


mapyr:
## Background

Observed on a live gateway running 2.19.1, while investigating #1356. Reproduces on every replica.

## 1. A group's indicator renders as `[?]`

`hangar_status` documents its indicator vocabulary as `[READY] [COLD] [STARTING] [DEGRADED] [DEAD]`. A group's state is `healthy`, which is not in that set, so the formatter falls through to `[?]`:

```
│ [COLD] weather-nws     cold     Will start on first   │
│ [?] weather         healthy  3/3 healthy           │
```

Servers and groups carry different state vocabularies. The human-readable surface is where that shows up, but the mismatch is upstream of the formatter — `hangar_list` and `hangar_details` return `state: "healthy"` for a group and `state: "cold" | "ready" | "degraded" | "dead"` for a server, and any consumer mapping one onto the other hits the same gap.

Decide which it is before patching the renderer: either groups gain the server vocabulary, or the two vocabularies are documented as distinct and the formatter gets a mapping for the second. Adding a `[HEALTHY]` case to the renderer without answering that just moves the inconsistency one layer down. Related in kind to #1361, which asks what the server state enum's own values mean.

## 2. The ASCII frame does not close

Every server row overruns the right border, and a member name is truncated mid-word:

```
╭─────────────────────────────────────────────────╮
│ MCP-Hangar Status                               │
├─────────────────────────────────────────────────┤
│ [COLD] github          cold     Will start on first   │
│ [READY] notes           ready                          │
│ [COLD] weather-open-me cold     Will start on first   │
│ [?] weather         healthy  3/3 healthy           │
├─────────────────────────────────────────────────┤
│ Health: 2/8 mcp_servers healthy                 │
╰─────────────────────────────────────────────────╯
```

**Correction to an earlier reading of this:** the first diagnosis here was that column widths are computed for one row shape and reused for another. That is wrong. A later capture with all eight rows in the same state shows every one of them overrunning by the same amount, so the row widths are simply computed independently of the frame — the border, the header and the footer agree with each other and the content rows agree with each other, and the two were never reconciled. The `[READY]` rows also sit one character wider than the `[COLD]` rows, so the indicator is being padded to its own length rather than to a column width.

`weather-open-meteo` is cut to `weather-open-me`, which is not a valid id and cannot be pasted into `hangar_details`.

## Acceptance criteria

[...]

Issue #1356, opened 2026-09-12
fix(core): report group healthy_count as members healthy, not members in rotation


mapyr:
## Background

Split out of #1315. Noticed while debugging the per-replica circuit breaker; it is what sent us looking in the wrong direction first.

## Why

`healthy_count` on a group reports members *in rotation*, not members *healthy*. A `cold` member counts. That produces self-contradictory snapshots an operator cannot act on:

```
healthy_count: 3, is_available: false, circuit_open: true
```

Three healthy members and an open circuit is not a state anyone can reason about, and both `GET /api/groups/{id}` and `hangar_details` report it.

## What changes

Make `healthy_count` mean what its name says, and expose rotation size as its own field rather than overloading one number with two meanings.

## Acceptance criteria

- [ ] `healthy_count` counts only members in `ready` state.
- [ ] Rotation membership is reported under its own field; the existing `members_in_rotation` list already carries the identities, so the count follows it rather than duplicating semantics.
- [ ] A test asserts the contradictory shape is no longer reachable: with all members `cold`, `healthy_count` is 0 while `members_in_rotation` may be non-empty.
- [ ] `GET /api/groups/{id}` and `hangar_details` agree field for field on the same group at the same instant.
- [ ] The field-name change is called out in the upgrade note if any name changes meaning rather than being added.

## Out of scope

Circuit breaker behaviour and rotation policy. This issue only fixes what the numbers report.

## References

- https://github.com/mcp-hangar/mcp-hangar/issues/1315

mapyr:
Reproduced on a live gateway running 2.19.0, from the opposite direction to the original report — and this shape is the more dangerous one.

`hangar_group_list`, 35 minutes after a pod restart, with zero calls made through the group:

```
group weather: state=healthy  healthy_count=3  total_members=3
               is_available=true  circuit_open=false
members: weather-open-meteo  state=cold  in_rotation=true  consecutive_failures=0
         weather-wttr        state=cold  in_rotation=true  consecutive_failures=0
         weather-nws         state=cold  in_rotation=true  consecutive_failures=0
```

Three cold members reported as three healthy ones. No failure has occurred; the group has simply never been called since the restart, so nothing has left rotation.

[...]

PR #1391, opened 2026-09-13
fix(core): name the replica that answers hangar_status and hangar_health

### LINKED ISSUE #1380: fix(core): hangar_status reports one replica's view as if it were the fleet
## Background

Observed on a live gateway running 2.19.1, during a deliberate load run.

## Why

Two calls to `hangar_status`, the same tool, minutes apart, returned incompatible pictures of the same fleet:

```
call 1:  uptime    34.8s   ->  healthy_mcp_servers 8 / 8,  every server [READY]
call 2:  uptime  4819.6s   ->  healthy_mcp_servers 2 / 8,  six servers [COLD]
```

35 seconds and 1h20m are not the same process. These are two replicas, each answering with its own local view, and the uptime counter is what makes it unambiguous — this is not the fleet changing between calls, it is two irreconcilable descriptions of one moment.

`hangar_status` is the first thing an operator reaches for. Under session affinity they do not choose which replica answers, so the answer is effectively arbitrary. An operator who polls it twice and sees the fleet "recover" or "collapse" is reading replica identity, not fleet state.

This is #1315's finding generalised past groups. There the symptom was that `GET /api/groups/{id}` answers differently depending on the pod. Here it is the whole server fleet, through the primary diagnostic tool. Same family as #877 (sessions are per-replica) and #1358 (rotation and circuit state are per-replica).

## A second contradiction, inside one response

The same payload from call 2 reports:

```
[COLD] weather-wttr    cold
[COLD] weather-nws     cold
[?]    weather         healthy   3/3 healthy
```

Two of the three group members are cold, and the group reports three healthy members, in one render. That is #1356, reproduced without having to line up two calls.

## What changes

Decide what `hangar_status` is allowed to claim, then make it claim that.

Three shapes, and the decision belongs to a maintainer:

1. **Fleet-wide truth** — status reads from shared state, so any replica answers identically. Consistent with #1358 if that lands the same way, and the two should be decided together rather than drifting apart.
2. **Honest replica-local** — the response names the replica and says plainly that it describes what this pod knows. Cheap, and at least stops the reader inferring a fleet fact from a pod fact.
3. **Aggregated** — status queries peers and reports the union, with disagreement visible rather than hidden behind whichever pod answered.

Whatever wins, `uptime` in a replica-local response is actively misleading as a fleet property.

## Acceptance criteria

- [ ] The decision is recorded in this issue before implementation, together with its relationship to #1358 — one answer, not two.
- [ ] A test asserts that two replicas with divergent local state produce responses an operator cannot mistake for a fleet-wide claim: either identical, or explicitly scoped to the replica.
- [ ] `hangar_health` and `hangar_status` agree on the same fleet at the same instant, from the same replica — today they are separate implementations of the same question.
- [ ] Whatever the outcome, the operator-facing docs state which of the three shapes applies.

## Out of scope

`healthy_count` semantics — #1356. The group circuit state design — #1358.

## References

- #1315, #1356, #1358, #877

mapyr:
## Why

Closes #1380.

Two `hangar_status` calls a few minutes apart reported uptime 34.8s with 8/8 healthy, then 4819.6s with 2/8 healthy. They reached two replicas, and each answered with its own local view in a shape that read as the fleet's. The maintainer decision on the issue is **variant 2, honest replica-local**: name the replica that answered, say the answer covers only that replica, and make `hangar_health` answer from the same source. There is no aggregation and no shared state. The fleet view belongs to Prometheus, and #1358 is separate.

## What

- **Replica identity is reused, not invented.** Both tools report `current_instance_id()`, the id minted at bootstrap as `HANGAR_INSTANCE_LABEL` or the hostname, plus a per-process suffix. The same value is the management lease `holder`, the peer event tailer's own-append filter, `produced_by` on events, `service.instance.id` on traces, and `instance.instance_id` on `GET /system`.
- **One source for both tools.** A new module, `server/tools/replica_view.py`, provides `observe_replica()`. It reads this replica's configured servers (the repository), hot-loaded servers (the runtime store) and groups once, together with the replica's id and uptime. `hangar_status` and `hangar_health` both render from it. Before this, `hangar_status` went through `ListMcpServersQuery` and counted hot-loaded servers, while `hangar_health` read the repository and left them out. From one replica, the two tools could report different totals.
- **Why the repository and not the query bus.** `ListMcpServersHandler` reads that same repository. But a probe of the served app that `build_serving_mcp_server()` builds, without a full `bootstrap()`, showed `hangar_status` already answering `HandlerNotRegisteredError` there. Routing `hangar_health` through the bus would have given a health check the same failure mode. Production registers the handlers first (`init_cqrs()` runs before `build_serving_mcp_server()` in `bootstrap()`), so the choice only matters where the wiring is partial, and there the health tool should still answer. The integration check below pins that.
- **Dead branches removed.** `hangar_status` had `hasattr(summary, "last_used_ago_s")` and `hasattr(summary, "consecutive_failures")` branches that `McpServerSummary` could never satisfy. They are removed, along with the `_format_time_ago` helper only they used. Output is unchanged.
- **New response fields (additive; no existing field renamed or removed)**, on both tools:
  - `replica: {instance_id, uptime_seconds, uptime}`
  - `scope: "replica"`
  - `scope_note: "This describes what the replica named in replica.instance_id knows, not the fleet. Other replicas can report different state and uptime at the same moment. For a fleet-wide view, query the per-replica metrics in Prometheus."`
  - `hangar_status.summary.uptime` / `uptime_seconds` stay, with the same values as `replica.uptime` / `replica.uptime_seconds`.
- **The rendered dashboard (`formatted`) names its replica:**

  ```text
  Answered by replica: gateway-b-3fa81c2e
  This replica's own view, not the fleet. Other replicas can differ.
  ╭─────────────────────────────────────────────────╮
  │ MCP-Hangar Status (this replica)                │
  ├─────────────────────────────────────────────────┤
  ...server and group rows unchanged...
  ├─────────────────────────────────────────────────┤
  │ Health: 2/8 mcp_servers healthy                 │
  │ Replica uptime: 1h 20m                          │
  ╰─────────────────────────────────────────────────╯
  ```

  The id sits above the frame so it is never truncated, because the suffix is what tells two replicas apart. Frame width, group rows and the `[?]` indicator are untouched; those are #1378's.
- **Tool descriptions** for `hangar_status` and `hangar_health` gain a `SCOPE: replica-local` paragraph and document the new fields.
- **Behaviour change:** `hangar_health.mcp_servers.total` and `by_state` now count hot-loaded servers, as `hangar_status` always has. That is the agreement AC3 asks for.
- **Not touched:** `healthy_count` semantics (#1356), `domain/model/mcp_server_group.py` and `application/sagas/group_rebalance_saga.py` (#1383), and `GET /system`, which already reports `instance.instance_id`.

## How tested

New `tests/unit/test_hangar_status_names_its_replica.py`:

- `TestTwoReplicasCannotBeReadAsOneFleet` replays the observed pair: replica `gateway-a` (8/8 ready, uptime 34.8s) and replica `gateway-b` (2/8 ready, uptime 4819.6s). Both tools on both replicas answer `scope: "replica"`, and each names its own distinct `instance_id`, equal to `current_instance_id()` at the moment it answered. Uptime is attributed to the replica, and the legacy summary fields equal `replica.*`. The dashboard's first line names the replica, and no bare `Uptime:` row remains. The two changed rows keep the frame width.
- `TestHealthAndStatusAgree` sets up one replica with ready, cold and degraded configured servers, a hot-loaded ready server and a group. The two tools report the same replica, the same server total (4), the same ready count (2), the same group totals and the same hot-loaded server.
- A wire-level check in `tests/integration/test_serve_modern_surface.py` calls both tools over the served modern streamable-HTTP app and asserts that the new fields survive serialization.

Local gates (private venv, `.[dev,opentelemetry]`, `mcp_hangar.__file__` inside this worktree):

| Gate | Command | Result |
| --- | --- | --- |
| lint | `ruff check src/mcp_hangar` (+ touched tests) | rc=0 |
| format | `ruff format --check src/mcp_hangar` (+ touched tests) | rc=0 |
| types | `mypy src/mcp_hangar` | rc=0, no issues in 430 files |
| layering | `lint-imports --config .importlinter` | rc=0, 1 kept, 0 broken |
| dead symbols | `python scripts/check_dead_symbols.py` | rc=0, baseline holds |
| changelog | `make changelog-check` | rc=0, 1 fragment valid |
| unit | `pytest --ignore=tests/integration` | 8004 passed, 63 skipped, **1 failed** (see below) |
| integration | `pytest tests/integration/ --timeout=60` | rc=0, 382 passed, 5 skipped |

The one unit failure is `tests/ci/test_base_images_are_pinned_and_updated.py::test_there_are_dockerfiles_to_check`. It fails with `assert 0 >= 4` (`where 0 = len([])`), because the environment finds no Dockerfiles in a `.claude/worktrees` checkout. It always fails there and is unrelated to this change. An earlier run also failed `tests/benchmark/test_proxy_overhead.py::TestFullProxyPath::test_full_path_p99_assertion`, but only because the integration suite was running in parallel. It passed in the final full run and passes alone.

## Docs follow-up

The operator-facing docs live in `mcp-hangar/docs` and are not edited here. Proposed wording for the `hangar_status` / `hangar_health` reference and the HA/replicas page:

> **Scope: replica-local.** With more than one gateway replica, `hangar_status` and `hangar_health` describe the replica that answered the call, not the fleet. Each response names that replica in `replica.instance_id` and states `scope: "replica"`. Uptime (`replica.uptime_seconds`, and `summary.uptime_seconds` on `hangar_status`) is that replica's process uptime.
>
> Under session affinity you do not choose which replica answers. Two calls can reach two replicas and report different server states and uptimes without anything in the fleet having changed. Compare `replica.instance_id` before reading a difference as a change.
>
> `replica.instance_id` is the same identity the management lease reports as `holder`, `GET /system` reports as `instance.instance_id`, and traces carry as `service.instance.id`. Set `HANGAR_INSTANCE_LABEL` from the pod name (downward API `metadata.name`) to make it recognisable. A per-process suffix is always appended.
>
> From the same replica, `hangar_status` and `hangar_health` read one snapshot and agree: `hangar_health.mcp_servers.total` equals `hangar_status.summary.total_mcp_servers`, and both count hot-loaded servers.
>
> Neither tool queries peers or shared state. For a fleet-wide view, query the per-replica metrics in Prometheus, which scrapes every replica. For example, compare `max by (...)` with `min by (...)` over a per-replica series to see whether the replicas agree.

## Risk and rollback

Low risk. The JSON change is additive. The one semantic change is `hangar_health` now counting hot-loaded servers. The rendered text changed its header and uptime label, so anything parsing `formatted` by exact text would notice; `formatted` is documented as a human-readable dashboard. Revert the single squash commit to roll back.

## Agent metadata

- [x] LOC declared upfront: +587 / -153 across 6 files (234 of the additions are tests: 207 unit, 27 integration; 14 are the changelog fragment)


mapyr:

- **Replica identity.** It reuses `current_instance_id()`, the same id the lease reports as `holder` and that events and traces carry, so one replica has one name everywhere.
- **Tool names are unchanged.** Registration names them explicitly (`@mcp.tool(name="hangar_status")` / `name="hangar_health"`). The `_hangar_*` names are only the Python functions.
- **Reading the repository directly loses no filtering.** `ListMcpServersHandler` reads the same repository and filters only by state, and `hangar_status` passed `state_filter=None`.
- **One snapshot for both tools.** Both render from `observe_replica()`, which closes the hot-loaded gap that let them disagree (AC3).
- **Nothing removed.** Response fields are only added.

CI is fully green. The docs follow-up (AC4) goes to mcp-hangar/docs after this merges, so the docs do not describe fields the released version lacks.

PR #1393, opened 2026-09-13
test(core): fail the build when a security predicate is unreachable from the request path

### LINKED ISSUE #1386: test(core): fail the build when a security predicate is unreachable from the request path
## Goal

CI fails when a security-relevant predicate exists but no served request path calls it.

## Motivation

Two recent defects share one pattern: code that exists, looks as if it does its job, and nothing calls or checks it.

- **`is_session_suspended()`** looked like a suspension check. No request path called it, so a suspended session kept invoking tools until 2.19.1 (GHSA-fhwh-fmq2-7m5c).
- **`PROJECTED_TOOLS`** looked like a metric, and nothing observed it (#1059). Dashboards built on it stayed empty.

Patches close the two cases. A test closes the class.

The repo already has the cure for one instance. `tests/unit/test_tool_permissions_cover_the_surface.py` reads the tool surface off the served server, and fails the build when a tool has no permission entry.

The dead-symbol ratchet (`scripts/check_dead_symbols.py`) does not cover this class. It counts a symbol as used when its name appears anywhere in `src/` or `tests/`, so a predicate referenced only by its own unit tests, or only by a path nothing serves, passes it.

## Proposed change

1. **Name the predicates in one place.** Keep a registry of security-relevant predicates, as a table the way `TOOL_PERMISSIONS` is one, or as a decorator. Each entry names the served request paths it must be reachable from.
   - Candidates for the first set: the session-suspension guard, `authorize_tool`, `_authorize_calls`, the tool-access resolver's allow decision, `is_governed_allowed`, the executor's withdrawal, digest-pin and approval gates, the L7 header evaluator, and the task store's `authorize`.
   - The implementer finalizes the list from the code.
2. **A reachability test that drives the served app.** It builds what `mcp-hangar serve` serves (`build_serving_mcp_server` / `mcp_app_for_serving`), not hand-assembled objects. That lesson comes from #596 and #963-#965.
   - It wraps each registered predicate in a spy.
   - It sends one request down each declared path: `hangar_call`, a flat `tools/call`, a `hangar_*` management tool, `tasks/*`, and the prompt and resource relays.
   - It fails when a declared path never invokes its predicate.
3. **Keep the registry from going stale.** Add a static check that flags every function in `src/` whose name has the shape of a predicate (for example `is_*_allowed`, `is_*_suspended`, `authorize*`, `refuse_if_*`, `check_*permission*`). Each one must be registered or explicitly exempted with a reason, the same ratchet style as the dead-symbol and complexity baselines.

## Acceptance criteria

- [ ] Removing the call to any registered predicate from one of its declared paths fails the reachability test. The PR shows this for each predicate with a mutation run.
- [ ] A new function with a predicate-shaped name that is neither registered nor exempted fails the static check.
- [ ] The reachability test drives the served app, not objects built by hand.
- [ ] Both checks run in CI on every PR.

## Out of scope

- Fixing a predicate the new test finds unreachable. Each such finding gets its own issue, or a private security advisory if it is a vulnerability.
- Metrics that are defined but never written. They are a sibling of this class and are tracked separately.

mapyr:
## Why

Closes #1386.

Twice a check existed, read as if it did its job, and nothing called it. `is_session_suspended()` refused no request until 2.19.1 (GHSA-fhwh-fmq2-7m5c), and nothing observed `PROJECTED_TOOLS` (#1059). The dead-symbol ratchet counts any reference in `src/` or `tests/` as a use, so it cannot see this class of defect. This PR adds a test that fails the build when a registered security predicate does not run on a served path it guards, and a static check that keeps the registry from going stale.

## What

- **`tests/unit/_security_predicates.py`: one table.** It lists 18 predicates, the served paths each must run on, the name patterns and the exemptions.
  - Where one request can reach a predicate from two call sites, the entry names the argument that tells them apart (`by=`). Examples are the session guard's `chokepoint` and `is_governed_allowed`'s `kind`.
  - Without that, deleting the `kind="prompt"` governance check would go unnoticed, because the flat tool map calls the same function with `kind="tool"` on the same request.
- **`test_security_predicates_are_reachable.py`: drives the served app.** It builds what `serve --http` serves, in both `egress` and `front_door`:
  - `build_serving_mcp_server()`, the task relay `bootstrap()` adds (`enable_governed_task_relay`), `mcp_app_for_serving()` and `create_auth_enforced_app`.
  - It sends one real streamable-HTTP request down each of 13 paths through `TestClient`, with a real API key, from loopback (a trusted proxy by default) and with an `x-session-id`.
  - The one upstream is an in-process HTTP MCP server. It serves tools, prompts, completions, resources and templates, and hands back a task handle for `tasks/*`.
  - It uses no mocks: no handler, gate or predicate is patched.
  - It records calls with a profile hook matched on each predicate's code object. That sees a call however it is reached: an import alias, the executor's `_GATES` tuple, a closure, or a worker thread.
- **`test_security_predicates_are_registered.py`: static check.** It flags every function in `src/` whose name has a predicate's shape: module functions, methods and nested functions. Each one must be registered, or exempt with a one-line reason. Stale exemptions and entries that are both registered and exempt also fail it.

### The registry

| Predicate | Declared paths |
| --- | --- |
| `session_guard.refuse_if_session_suspended` (by `chokepoint`) | all 13 |
| `session_guard.refuse_request_if_session_suspended` (by `chokepoint`) | `tasks/*`, prompts, completion, resources |
| `InMemorySessionSuspensionRegistry.is_suspended` | all 13 |
| `tool_permissions.authorize_tool` | `hangar_*` tool, front-door `hangar_*` tool |
| `batch._authorize_calls` | `hangar_call` |
| `AuthorizationMiddleware.authorize` | `hangar_call`, `hangar_*` tool, front-door `hangar_*` tool |
| `ToolAccessResolver.is_tool_allowed` | `hangar_call`, front-door `tools/call` |
| `ToolAccessResolver.is_allowed` (by `kind`) | `hangar_call`, front-door `tools/call`, prompts, completion, resources |
| `flat_tool_projection.is_governed_allowed` (by `kind`) | front-door `tools/call`, prompts, completion, resources |
| `ToolProjectionRegistry.is_withdrawn` (by `kind`) | front-door `tools/call`, prompts, completion, resources |
| `ToolProjection.is_withdrawn_for` (the executor's withdrawal gate) | `hangar_call`, front-door `tools/call` |
| `BatchExecutor._enforce_digest_pin` | `hangar_call`, front-door `tools/call` |
| `BatchExecutor._check_approval_gate` | `hangar_call`, front-door `tools/call` |
| `McpServer._enforce_l7_policy` | `hangar_call`, front-door `tools/call` |
| `egress_l7.evaluate_headers` | `hangar_call`, front-door `tools/call` |
| `GovernedTaskStore.authorize` | `tasks/get`, `tasks/cancel`, `tasks/update` |
| `UiResourceGuard.evaluate` | `resources/list`, `resources/templates/list`, `resources/read` |
| `UiResourceGuard.enforce` | `resources/read` |

The static check uses these patterns:

- `^_?is_\w*(allowed|permitted|authori[sz]ed|suspended|denied|blocked|withdrawn)\w*$`
- `^_?authori[sz]e\w*$`
- `^_?refuse(_\w+)?_if_\w+$`
- `^_?check_\w*(permission|access|approval|auth)\w*$`
- `^_?enforce(_\w+)?$`

They flag 38 functions today: 16 registered and 22 exempt. The exemptions fall into three groups:

- The rule behind a registered predicate: value-object pattern matches, the ownership rule, the overlay internals, and the hook runner.
- Authorizer backends and interfaces.
- Functions not on an MCP request path: REST guards and routes, the startup reachability check, and JWT lifetime and audience checks, which the probes don't reach because they authenticate with API keys.

`subscriptions/listen` has no probe. It is a long-lived stream, and the issue's path list does not include it.

## How tested

**Mutation proof, one per predicate.** For each registered predicate, I deleted or bypassed its call on one declared path, ran `tests/unit/test_security_predicates_are_reachable.py` in a fresh interpreter, and restored the file. All 18 failed. The request probes stayed served each time, so the failures are reachability failures, not broken probes.

| Predicate | Mutation (file) | Result |
| --- | --- | --- |
| `refuse_if_session_suspended` | delete the guard call at `hangar_call`'s entry (`batch/__init__.py`) | 2 failed: guard and registry lookup on `hangar_call` |
| `refuse_request_if_session_suspended` | delete it from the `prompts/list` handler (`prompt_proxy.py`) | 3 failed on `prompts/list` |
| `InMemorySessionSuspensionRegistry.is_suspended` | `suspended = False` inside the guard (`session_guard.py`) | 13 failed, one per path |
| `authorize_tool` | stop `register_all_tools` installing the authorizer (`bootstrap/tools.py`) | 6 failed on both `hangar_*` tool paths |
| `_authorize_calls` | `denied_by_index = {}` in `hangar_call` | 2 failed on `hangar_call` |
| `AuthorizationMiddleware.authorize` | bypass `authz.authorize` inside `_authorize_calls` | 1 failed on `hangar_call` |
| `ToolAccessResolver.is_tool_allowed` | drop `_gate_tool_access` from `_GATES` (`executor.py`) | 3 failed on `hangar_call` and front-door `tools/call` |
| `ToolAccessResolver.is_allowed` | `return True or self.is_allowed(...)` in `is_tool_allowed` | 1 failed on `hangar_call` |
| `is_governed_allowed` | delete the `kind="prompt"` check in `_build_prompt_map` | 9 failed on prompts/completion (kind `prompt`) |
| `ToolProjectionRegistry.is_withdrawn` | delete the withdrawal half of `is_governed_allowed` | 7 failed, all its paths |
| `ToolProjection.is_withdrawn_for` | drop `_gate_withdrawal` from `_GATES` | 1 failed on `hangar_call` |
| `BatchExecutor._enforce_digest_pin` | drop `_gate_digest_pin` from `_GATES` | 2 failed on both invoke paths |
| `BatchExecutor._check_approval_gate` | drop `_gate_approval` from `_GATES` | 2 failed on both invoke paths |
| `McpServer._enforce_l7_policy` | delete the call from `McpServer.invoke_tool` | 2 failed on both invoke paths |
| `evaluate_headers` | `header_verdict = None` inside `egress_l7.evaluate` | 2 failed on both invoke paths |
| `GovernedTaskStore.authorize` | delete the ownership check from the `tasks/cancel` handler | 1 failed on `tasks/cancel` |
| `UiResourceGuard.evaluate` | delete the `ui://` check from `_deliverable` | 2 failed on the two resource listings |
| `UiResourceGuard.enforce` | skip consent enforcement in `resources/read` | 1 failed on `resources/read` |

**Static check.** Appending a scratch `def is_scratch_allowed()` to `server/session_guard.py` fails `test_it_is_registered_or_exempt`. The scanner's own tests pin the same thing in a temporary tree.

**Gates, in a private venv (`mcp_hangar.__file__` inside the worktree):**

- The two new files: 112 passed on Python 3.11, 3.12, 3.13 and 3.14. Each run takes about 3 to 6 seconds.
- `ruff check` and `ruff format --check` (`src/` and the new files), `mypy src/mcp_hangar`, `lint-imports` and `scripts/check_dead_symbols.py`: all clean.
- `pytest --ignore=tests/integration` (CI's unit command): 8103 passed, 63 skipped, 1 failed. The one failure is `tests/ci/test_base_images_are_pinned_and_updated.py::test_there_are_dockerfiles_to_check`. It always fails under `.claude/worktrees` and is environmental, not related to this change.
- `pytest tests/integration/ --timeout=60`: 380 passed, 5 skipped.

Both new files are in `tests/unit`, so `ci-core`'s `test` job runs them on every PR. It has no path filter on `pull_request`.

## Risk and rollback

This changes tests only. The fixture resets every process-global it touches: context, resolver topology, projection registry, the installed tool authorizer, the task-relay flag and the trusted-proxy cache. It also restores the previous profile hooks.

A future change that moves a predicate call behind a path the probes don't take will fail here. That is the point: either fix the wiring or correct the table. Rollback is a single revert.

## Agent metadata

- [x] LOC declared upfront: 923 (tests only)


mapyr:
Reviewed and ready.

- **Registry.** One table, like `TOOL_PERMISSIONS`, lists 18 predicates, each with the served paths it must run on. `by` pins the call site, so dropping only the `kind="prompt"` governance check is caught even though `kind="tool"` runs on the same request.
- **Reachability.** The test drives the app `serve --http` builds, in `egress` and `front_door`, over real streamable HTTP and with real API-key auth. It patches no handler, gate or predicate. Every one of the 18 predicates fails the test when its call is removed from a declared path; the per-predicate results are in the PR body.
- **Spy.** It uses `sys.setprofile`/`threading.setprofile`, a separate hook from coverage's tracer, so it coexists with `decision-coverage`. If it misses something, such as a pool thread started before the hook, the test fails loudly rather than passing falsely.
- **Static ratchet.** It flags every function in `src/` with a predicate-shaped name. That is 38 today: 16 registered and 22 exempt, each with a reason. It also fails on stale exemptions and missing reasons.

CI is fully green.

Issue #1362, opened 2026-09-12
feat(core): log one structured line per call on the front door's projected surface


mapyr:
## Background

Split out of #1307 (first of its two gaps). Observed on core 2.18.2, front door with 5 replicas, `tool_access.mode: front_door`, OIDC tenants via Keycloak.

## Why

The front door logs nothing per projected tool call:

```
projected read_graph calls made: 5, succeeded: 5

kubectl -n frontdoor logs -l app.kubernetes.io/name=mcp-hangar --since=3m | grep -c read_graph
0
```

Five successful calls, zero lines naming the tool, one generic event across all of them. Calls routed through `hangar_call` do produce a `BatchCallCompleted` domain event; calls to a flat projected tool produce nothing that identifies what was called.

The cost is concrete: a client reported a deterministic failure of `read_graph` and `search_nodes("")` carrying its own request id, and there was nothing server-side to correlate it with — no request id, no tool name, no error, no verdict. Every reproduction attempt (empty `arguments`, missing `arguments`, `arguments: null`, empty and non-empty graph, repeated within and across sessions) returned 200. The failure is still unexplained and the log gap is why it will stay that way.

## What changes

One structured line per tool call on the projected surface, at the same level the meta-API's own calls already get.

## Acceptance criteria

- [ ] Every call through the flat projected surface emits one structured line carrying tool name, principal, tenant, outcome and duration.
- [ ] A denied call emits a line too, carrying the verdict — a deny that leaves no trace is the case this issue exists for.
- [ ] A test drives N calls through the projected surface and asserts N lines naming the tool.
- [ ] Arguments are not logged raw; redaction follows `domain/security/argument_redaction.py`, which already exists for exactly this class of leak.
- [ ] Log volume is bounded on the hot path: one line per call, no per-argument expansion.

## Out of scope

`hangar_whoami` and identity introspection — the second half of #1307, split out separately because it adds a public tool surface and needs a permissions decision.

## References

- https://github.com/mcp-hangar/mcp-hangar/issues/1307

Issue #905, opened 2026-08-11
proposal: the shape of a policy denial for an authenticated principal is indistinguishable from a missing tool


mapyr:
Verified against the published **2.5.3** wheel from PyPI (not `main`). **Proposal only — do not implement from this issue.** It changes the protocol contract a client sees and deserves its own cycle.

A tool denied by policy comes back as one flat refusal, whoever asked:

```python
# server/tools/batch/executor.py:1060
return p.refuse("Tool not available for this mcp_server", "ToolAccessDeniedError")
```

There is no branch on caller identity anywhere above it — `_gate_tool_access` computes `allowed` and refuses (`executor.py:1026-1060`).

Against an anonymous caller that is the right answer: it leaks nothing about what exists. Against a **registered, authenticated principal** it is indistinguishable from "there is no such tool", which is also what an upstream failure looks like. The agent has nothing to reason from and usually retries the same call.

Two things narrow the scope of any proposal here.

**Differentiating by reason already exists**, 23 lines further down — the withdrawal gate says which tenant and why:

```python
# executor.py:1083
return p.refuse(f"Tool '{p.call.tool}' is withdrawn for this tenant", "ToolWithdrawnError")
```

So the missing axis is identity, not reason. That is a smaller change than it first looks.

**The envelope is the harder half.** The richer shape the task path uses —

```python
# tasks_wire.py:92, 243-250
MISSING_REQUIRED_CLIENT_CAPABILITY = -32021
def _required_capabilities_payload():
    return {"requiredCapabilities": {"extensions": {EXTENSION_ID: {}}}}
```

— is a JSON-RPC error. On the default `egress` path a denial is not a JSON-RPC error at all: it is `isError=True` inside the `hangar_call` batch envelope, one entry among many in a batch that may otherwise have succeeded. Carrying `-32021`-shaped structure there is a decision about the batch result schema, not just about an error code. On `front_door` the picture differs again — an unknown flat name is `-32601` and a denial is a `CallToolResult(isError=True)` (`fastmcp_server/flat_tool_projection.py:399-435`).

## The question to settle

1. Should the refusal shape differ by whether the caller is an authenticated principal or anonymous? (The information-disclosure argument for the flat message applies only to the anonymous case.)
2. If yes, what does the authenticated caller learn — that the tool exists but policy denies it, which policy surface denied it, whether an approval path exists?
3. How is that carried on the `egress` batch envelope, which is where the default deployment lives?

[...]

Issue #877, opened 2026-08-11
Streamable HTTP sessions are per-replica, so a coordinated deployment is not one server to a client


mapyr:
A Streamable HTTP session lives in one replica's memory. Three replicas of one
coordinated gateway — one database, one management lease, one identity by
ADR-020 — are therefore three different servers to any MCP client, and the
Service in front of them hands each request to whichever one it feels like.

Measured on core `2.5.2`, `replicaCount: 3`, through the chart's own ClusterIP
Service. Each attempt is one client doing `initialize` → `tools/list` →
session teardown:

```
15 attempts through the Service (3 replicas):
   1  clean
   1  request served, teardown 404 (the session leaks on the owning replica)
  13  the request itself failed: "Session not found"
```

The only variable is the replica count:

```
SAME Service, replicaCount=1: 10 ok / 0 failed
SAME Service, replicaCount=3:  6 ok / 4 failed
```

(The rate moves with connection reuse — a client that keeps one TCP connection
for the whole session gets lucky. Anything that opens a second connection, and
anything behind a proxy that pools, does not.)

Against a single pod directly: 12 of 12 clean.

## Why this is a core issue and not only a chart one

The chart can paper over it — `sessionAffinity: ClientIP`, or an Ingress that
hashes on the client address, which is what we ended up doing in our lab and is
why we did not see this for weeks. Both are per-deployment mitigations for
something the product asserts is not true: that these replicas are one gateway.

They are one gateway for the lease, for configuration, for the event log and for
the audit trail. They are not one gateway for the protocol the product exists to
speak. A client that reconnects, a proxy that pools, a rolling restart, a
scale-up — each is a session lost, with `Session not found` and no path to
recovery except starting over.

## What a fix could look like

Not obvious, and worth an ADR rather than a patch:

1. **Shared session state** — the session id and its state in PostgreSQL, so any
   replica can serve any session. Truthful, and the most work; session state is
   not currently written down anywhere.
2. **Session affinity as a documented deployment requirement** — the honest
   short-term answer, but then the HA docs must say plainly that MCP clients
   require sticky routing, and the chart must make it reachable (see the
   helm-charts issue linked below).
3. **Redirect on miss** — a replica that does not own a session answers with
   something the client can act on rather than a flat error.

Filing this as the statement of the problem; the chart-side mitigation is
tracked separately.

mapyr:
[...]

PR #1395, opened 2026-09-13
feat(core): a call to a name that left the projection returns -32601 with a staleness reason

### LINKED ISSUE #1368: feat(core): a call to a name that left the projection returns -32601 with a staleness reason
## Background

Part of #1365.

Once the projection can change during a session, a client holding a stale list will call a name that is no longer projected. Today that is indistinguishable from calling a name that never existed, so a client's only strategy is to retry or give up.

## What changes

A call to a name that was projected for this caller and no longer is returns `-32601` carrying a reason that tells the client its list is stale, so it re-lists instead of retrying.

## The constraint this must respect

The project has decided, in #905, that a denial must not leak whether a thing exists — the indistinguishability of "denied" from "absent" is load-bearing against enumeration of another tenant's surface, and that issue was closed working-as-intended on exactly that ground.

So the reason must say "your list is out of date", never "this tool exists but you may not call it" and never "this tool was withdrawn from you specifically". A caller who never had the name projected gets the ordinary `-32601` with no reason attached.

## Acceptance criteria

- [ ] A name that left this caller's projection returns `-32601` with a machine-readable reason indicating the projection changed.
- [ ] A name that was never projected for this caller returns the ordinary `-32601`, with no reason — verified by a test that a caller cannot distinguish "never existed" from "exists but not for you".
- [ ] The reason carries no upstream id, tool name from another tenant, or count.
- [ ] A test drives the full sequence: list, projection changes, call the stale name, receive the reason, re-list, call succeeds.

## Out of scope

The notification that tells the client to re-list — #1366. Whether the client acts on the reason.

## References

- #905

mapyr:
## Why

Closes #1368. Part of #1365.

Once the projection can change under a connected client, a client holding a stale list calls a name that has left it. Until now that got the same `-32601` as a name that never existed, so the client could only retry or give up. Now the one case that can be answered honestly, "this caller was served the name and no longer is", says so, and the client lists again. Everything #905 decided stays true: a caller never learns that a name exists for someone else.

## What

- A front-door `tools/call` to a name that the caller's last `tools/list` on this replica served, and that is no longer in its projection, answers:

  ```json
  {"code": -32601, "message": "Tool 'write_item' not found", "data": {"reason": "projection_changed"}}
  ```

  The code and the message are unchanged, and the reason lives only in `data`, so a client that does not know it reads today's error. The payload follows the `tasks_wire.missing_capability_error_data` convention: the code says what failed, and `data` says what to do about it.
- Any other unknown name gets the ordinary `-32601` with no `data`, byte for byte as before. That covers a name held by another tenant, a name policy denies this caller, and a name that exists nowhere. Existence is never consulted. Only what this caller was served is.
- The reason is a constant. It carries no upstream id, no tool name and no count, and it does not say why the name left: a withdrawal for this tenant, a withdrawal for everyone, a policy edit and a vanished upstream all read the same.
- New `fastmcp_server/served_tool_names.py` holds the memory:
  - **What:** the names in the last real `tools/list` answered for an identity, governed and management tools alike. An empty listing clears the entry.
  - **Key:** tenant plus principal (type, user, agent), narrowed by the verified session (`CallerIdentity.session_id`: the token's `sid` or a trusted proxy's `x-session-id`, the one suspension keys on). The tenant and the principal are always in the key, so a session narrows an entry and never widens it. `Mcp-Session-Id` is not used: the served app is stateless (#877), and 2026-07-28 has no sessions. Without a session, two clients of one principal share an entry and the later listing wins. That can only give that same principal a reason for a name it was served. A caller without a tenant is never remembered.
  - **Timing:** recorded only for a real listing. The SDK's pre-dispatch `tools/list` on a `tools/call` (#1049) is not a listing the client received and is not recorded. It reuses the branch that already keeps that listing out of `PROJECTED_TOOLS`.
  - **Bound:** a thread-safe LRU of 1024 identities, holding name sets only. An evicted identity gets the ordinary `-32601`.
  - **Per replica:** a call that reaches a replica that did not serve the list gets the ordinary `-32601`. Both failure modes fall back to today's behaviour, and neither can produce a reason for a name the caller was not served.
- `flat_tool_projection.py` changes in three places. One line records the listing inside `_list_projected_tools`' existing real-listing branch. There is a new `_not_found_error(name)` helper. The `-32601` raise in `_flat_call_tool` becomes `raise _not_found_error(name)`. Nothing else in the call handler moves, so the per-call logging in #1362 should meet no conflict. `register_flat_tool_handlers` stays at complexity 15.

## How tested

The served path, over the real streamable-HTTP transport (`tests/integration/test_a_stale_name_is_told_to_list_again.py`):

- **Setup:** the composition `serve --http` serves, `mcp_app_for_serving(build_serving_mcp_server())`, behind `create_auth_enforced_app`, with one API key per tenant.
- **Upstream:** one real HTTP upstream behind a real `McpServer`, started through the command bus. Its catalogue reaches the registry through the `McpServerStarted` handler, subscribed as bootstrap does.
- **Projection changes:** made through the real registry (a withdrawal) and the real resolver (a policy edit).

What the integration file tests:

- **Full sequence:** list, a policy edit changes the projection, call the stale name and get the reason, list again, call a name from the fresh list, and it reaches the upstream and succeeds.
- **Really gone:** after listing again, the withdrawn name gets the ordinary `-32601`, with no `data`.
- **Handshake era:** the same on the 2025-06-18 era, which answers through the front-door wrap and the SSE framing.
- **Indistinguishability proof:** tenant A holds `write_item` and has it withdrawn. Tenant B is denied it by policy and was never served it. B's answer for `write_item` is compared with B's answer in a second fleet, with the same configuration and history, where `write_item` exists for nobody. The status, content type and raw body bytes are identical. B's error equals the ordinary error, as does B's error for a never-existing name. A's error is the same object plus `data`.
- **The reason carries nothing:** `data == {"reason": "projection_changed"}`, with no server id, no tool name, no tenant and no digit in it.
- **Pre-dispatch listing:** a 2026-07-28 call with arguments makes the SDK list first. A spy confirms that listing ran, and it is not remembered.

Unit tests (`tests/unit/test_a_name_that_left_the_projection_says_the_list_is_stale.py`) cover:

- the handlers with the real registry and resolver
- key isolation across tenant, principal and verified session
- the no-tenant caller
- an empty listing clearing the entry
- LRU eviction order
- a bound below one being refused
- 8 threads remembering and reading concurrently while the bound holds

Mutation-checked before opening. With `_not_found_error` changed to attach the reason whenever the name exists anywhere, 11 tests fail, including both indistinguishability proofs. Changed to never attach it, 9 fail.

Gates, run locally with explicit exit codes:

```text
ruff check src/mcp_hangar                    exit 0
ruff format --check src/mcp_hangar          exit 0
mypy src/mcp_hangar                          exit 0  (430 files)
lint-imports --config .importlinter          exit 0  (1 kept, 0 broken)
python scripts/check_dead_symbols.py         exit 0  (baseline holds)
ruff check src --select C901                 exit 0
pytest --ignore=tests/integration            exit 1  (8016 passed, 63 skipped, 1 failed: see below)
pytest tests/integration                     exit 0  (386 passed, 5 skipped)
```

The one unit failure is environmental: `tests/ci/test_base_images_are_pinned_and_updated.py::test_there_are_dockerfiles_to_check` (`assert 0 >= 4`) always fails when the checkout lives under `.claude/worktrees`, because the Dockerfile scan finds nothing there. It is unrelated to this change and passes in a normal checkout.

## Risk and rollback

Low. For any caller not served the name, the wire answer is unchanged by construction: `data` is None, and the SDK builds `ErrorData` without it. A client that ignores `data` sees today's error in every case. The memory is bounded and per process. Rollback: revert the squash commit.

## Agent metadata

- [x] LOC declared upfront: ~190 source (164 in the new module, +27/-2 in `flat_tool_projection.py`), ~665 test, 11 changelog


mapyr:
Reviewed and ready.

- **#905 holds.** `_not_found_error` asks only whether this caller's last listing on this replica served the name. It never consults the registry, so a name another tenant holds, a name policy denies, and a name that exists nowhere all get one answer. The integration test compares raw response bytes against a second fleet in which the tool does not exist.
- **Only listings the client received are remembered.** The SDK's pre-dispatch listing is excluded, and a caller with no tenant is never remembered. The key is tenant plus principal, narrowed by a verified session when there is one.
- **Bounded and per-replica.** The store is an LRU of 1024 identities and holds only name sets. On an eviction or another replica, the caller gets the ordinary `-32601`, which is today's behaviour.
- **Mutations are caught.** Attaching the reason too widely fails 11 tests, including both indistinguishability proofs.

CI is fully green. This should merge cleanly next to #1396, which touches only the handler's decorators and not the `-32601` branch.

PR #1396, opened 2026-09-13
feat(core): log one structured line per call on the front door's projected surface

### LINKED ISSUE #1362: feat(core): log one structured line per call on the front door's projected surface

mapyr:
## Why

The front door logged nothing per flat projected tool call that said who called what and how it ended. A client's deterministic failure, reported with its own request id, had nothing server-side to be matched against. A denied call left no trace at all. On this surface a deny is a `-32601`, or a suspended session's refusal.

Closes #1362
Part of #1365

## What

- Add `fastmcp_server/flat_call_log.py`. Its `logging_each_call` decorator, the outermost on the flat `tools/call` handler, writes exactly one INFO line, `front_door_tool_call`, per call, whatever the outcome.
- **Fields:**
  - `tool`, `outcome`, `reason`, `principal_id`, `tenant_id`, `request_id` (the JSON-RPC id) and `duration_ms`;
  - `trace_id` and `span_id` when tracing is on, added by the log pipeline from the server span the SDK opens around every handler.
- **Outcomes (six, bounded):**

  | outcome | when | `reason` |
  |---|---|---|
  | `ok` | served, result not an error | none |
  | `tool_error` | caller got `isError`, not a refusal: upstream failed or answered an error, timeout, cold start, open circuit | the executor's error type, if any |
  | `denied` | enforcement refusal: policy, withdrawal, digest pin, approval, validator, egress, suspended session | the refusal code, e.g. `ToolAccessDeniedError`, `session_suspended` |
  | `not_found` | `-32601` | `not_projected` (an upstream tool this gateway holds, not projected to this caller: the shape a policy deny takes here, #905) or `unknown` |
  | `rejected` | any other JSON-RPC error from the handler | `header_mismatch`, else `jsonrpc_<code>` |
  | `error` | unexpected exception | its class name |

- **No arguments are logged**: not raw, not redacted, not counted. I chose none over redaction via `argument_redaction.py`. The line exists to correlate a call, and request id, caller and tool do that. Every argument value is a potential secret that redaction would have to recognise. The two caller-chosen strings, the tool name and a string request id, are cut to 128 characters. One line per call with no per-argument expansion keeps the hot path bounded.
- **A suspended session's refusal is the one line without a tool name.** It carries neither the tool name nor the request id: nothing a suspended caller chose is echoed into the log. The session guard's own refusal line already follows that rule, and an existing test pins it. The principal, tenant and verdict are still on the line.
- **What the caller gets does not change.** The result or exception passes through untouched. The `not_projected`/`unknown` split exists only in the log, and computing it does the same work either way: one set built from the whole catalogue.
- **INFO**, the level a `hangar_call` call reaches the log at (`domain_event` for `BatchCallCompleted`).
- **`flat_tool_projection.py`, 6 lines added:**
  - the import and the decorator;
  - two `note_failure(...)` calls, one in the executor-refusal branch and one in the suspension decorator, so the line can carry the refusal code that the `isError` text does not.

  The `-32601` branch is untouched, so it doesn't conflict with the staleness-reason work (#1368). The complexity of `register_flat_tool_handlers` is unchanged, and the C901 gate passes.

## How tested

- **Served path**, in `tests/unit/test_every_front_door_call_leaves_one_log_line.py`. It drives `mcp_app_for_serving(build_serving_mcp_server())` over HTTP with `tool_access.mode: front_door` and a principal carrying a tenant, and reads what `setup_logging`'s JSON formatter renders. Only the auth middleware and the upstream (`BatchExecutor`) are stood in for. The tests cover:
  - N calls to one tool give N lines naming it, with the request ids in order;
  - one line per outcome, with its verdict: an executor refusal, a policy-filtered name, an unknown name, a timeout, an upstream `isError`, an exception, a suspended session and a header refusal;
  - a mixed run gives one line per call, and no argument appears anywhere Hangar logged at INFO;
  - denied and missing return the same status, code, message shape and `data` to the caller, while the log tells them apart;
  - a drift check: every error type the executor writes must be sorted into denial or not.
- **Live**, in `tests/live/test_t0_front_door_call_log.py`. It runs a real `mcp-hangar serve --http`, front door, with an API-key tenant and a per-tenant deny, called over streamable HTTP. Result: `1 passed`, one line per call, including the denied `power` (`not_found`/`not_projected`). A line from that run:

  ```json
  {"tool": "power", "outcome": "not_found", "reason": "not_projected", "principal_id": "svc:tenant-log", "tenant_id": "tenant-log", "request_id": 7, "duration_ms": 7.1, "event": "front_door_tool_call", "level": "info", "logger": "mcp_hangar.fastmcp_server.flat_call_log", "trace_id": "e0ee67d065d3474777ee84303bdfedf9", "span_id": "b180fe5fef354bf6", ...}
  ```

- **Gates:** `ruff check`, `ruff format --check`, `mypy src/mcp_hangar`, `lint-imports`, `scripts/check_dead_symbols.py` and `ruff check src --select C901` all pass.
- **Suites:**
  - `pytest --ignore=tests/integration` (the CI selection): 8014 passed, 64 skipped, 1 failed. The failure is `tests/ci/test_base_images_are_pinned_and_updated.py::test_there_are_dockerfiles_to_check`, which always fails under `.claude/worktrees` because the scan skips that path. It is environmental and not related to this change.
  - `pytest tests/integration`: 380 passed, 5 skipped.
- **The suspended-session rule was found by the existing suite.** `test_a_suspended_session_is_refused.py::TestTheGuard::test_the_log_line_names_only_the_chokepoint_and_the_session` failed on the first version, which named the tool on a suspended caller's line. This PR conforms to that test rather than changing it.

## Not covered

- **Refusals that happen before the `tools/call` handler do not produce this line.** Two cases:
  - an auth 401/403;
  - a legacy-era SEP-2243 header/body mismatch that `front_door_routing` rejects before dispatch. Its `_reject` logs nothing today.

  They are a different chokepoint and a follow-up.
- **#1395 (the #1368 staleness reason) composes with this PR.** It changes only the body of the `-32601` branch, which this PR leaves untouched. The `-32601` and its `data` pass through the wrapper unchanged. The log's `not_projected`/`unknown` reason comes from the catalogue and does not read `data`. Surfacing `projection_changed` on the line once #1395 lands is a small follow-up.
- **A call that reaches the executor produces two INFO lines.** It still publishes `BatchCallCompleted`, logged as a `domain_event` line. That line names the tool but not the caller.

## Risk and rollback

Low. One INFO line per front-door call, and the log call is fault-barriered, so it can never fail the call it describes. To roll back, revert the single commit.

## Agent metadata

- [x] LOC declared upfront: 231 source lines (225 in the new module, 6 in `flat_tool_projection.py`), 524 test lines (366 unit, 158 live), 9 lines of changelog


mapyr:
Reviewed and ready.

- **One line per call, whatever happens.** The decorator is outermost and passes the result or exception through untouched. `note_failure` recovers the refusal code that an `isError` result no longer carries. A logging failure can never fail the call.
- **No arguments are logged, not even redacted.** A suspended session's refusal echoes nothing the caller chose, which is the GHSA-fhwh rule, and the existing test for it caught the first version of this change.
- **What the caller receives is unchanged.** Denied and missing are still indistinguishable (#905). `not_projected` versus `unknown` exists only in the operator's log.

**One decision for the maintainer.** `principal_id` is new at INFO. No other INFO line on `main` carries a principal or user id. For OIDC users it can be an email-like `sub`, and logs often end up in pipelines outside the cluster. The issue asks for the principal on this line, so it stays raw here. The alternative is a stable hash on the log line, with the raw id kept on the audit record, which already carries caller identity by design.

CI is fully green.

Issue #1388, opened 2026-09-13
fix(core): group circuit state is never restored at startup, and would be restored to the wrong group


mapyr:
## Symptom

Persisted group circuit-breaker state is never restored when a gateway starts.

`init_saga()` calls `_restore_group_circuit_breakers(saga_state_store, ctx.groups)` (`server/bootstrap/cqrs.py:353`). At that point in `bootstrap()`, `ctx.groups` is still the empty dict that `init_context` created. The context is pointed at `GROUPS` only at the end of bootstrap. The loop therefore runs over nothing, and nothing is restored.

This is the same ordering defect that left `GroupRebalanceSaga` blind (#1355). PR #1383 fixes it for the saga at `cqrs.py:337`, but not for this restore.

## Root cause

There are two problems, and fixing only the first would make things worse:

1. **The empty dict.** The restore runs against `ctx.groups` before the context is pointed at `GROUPS`.
2. **The lookup is not keyed by group.** The docstring says breaker state is persisted "under `saga_type="circuit_breaker"` with `saga_id=group_id`", but the loop calls `store.load("circuit_breaker")` without a group. Once the loop runs over real groups, every group would get the same persisted row.

## What

Decide first whether startup should restore breaker state at all:

- Since #1383, a group's circuit closes on its own once members recover. Restoring a circuit opened hours ago could hold traffic off a group that is healthy now.
- State stays replica-local for now (the decision on #1380 and #1358).

Then do one of these:

- **Restore it correctly:** pass `GROUPS`, and load by group id.
- **Remove the dead path:** delete the restore and the persistence it reads.

## Files in scope

- `src/mcp_hangar/server/bootstrap/cqrs.py`
- Tests, under `tests/`.

## Acceptance criteria

- [ ] The decision is recorded here: restore or remove.
- [ ] If restoring: a test with two groups whose persisted states differ shows each is restored to its own state through the real `bootstrap()`, and a group with no persisted row starts closed.
- [ ] If removing: nothing writes state that nothing reads.

## Severity

P3: today it is dead code that restores nothing. A fix that makes it run without keying the lookup by group would open circuits on the wrong groups.

mapyr:
**Decision (maintainer, 2026-09-14): remove.**

Since #1383 a group's circuit closes by itself once its members recover. State stays replica-local (#1380, #1358). Restoring a circuit that was persisted as open hours earlier would hold traffic off a group that is healthy now.

[...]

PR #1397, opened 2026-09-13
fix(core): count a group's circuit failures in a row, not over the process lifetime

### LINKED ISSUE #1390: fix(core): a group circuit counts failures over the process lifetime, not in a row
## Symptom

A group's circuit opens after `circuit_failure_threshold` failures counted over the whole life of the process (default 10), not after that many *consecutive* failures. A healthy group that sees one member failure a day opens its circuit on day ten.

## Root cause

`CircuitBreaker` is written as a consecutive-failure breaker: in CLOSED state, `record_success()` resets the failure count (`domain/model/circuit_breaker.py:127-148`: "In CLOSED, resets failure count."). `McpServerGroup` never calls it while closed:

- **`report_failure()`** calls `self._circuit_breaker.record_failure()` (`domain/model/mcp_server_group.py:612`);
- **`report_success()`** never touches the breaker on `main`;
- the only other reset is `reset()`, called from `rebalance()` (`:693`), and only `hangar_group_rebalance` calls that.

So the count only grows until someone rebalances.

#1383 adds `record_success()`, but only to close an OPEN circuit once `min_healthy` members are in rotation (`_maybe_close_circuit`). The CLOSED-state reset is still never reached.

## What

After #1383 lands, have `report_success()` call `record_success()` when the circuit is not open, so a success resets the count as the breaker intends.

Leave the OPEN case to `_maybe_close_circuit()`: calling `record_success()` in OPEN closes the circuit at once, without its `min_healthy` check.

## Files in scope

- `src/mcp_hangar/domain/model/mcp_server_group.py`
- `tests/unit/`

## Acceptance criteria

- [ ] With `circuit_failure_threshold: 3`, the sequence failure, failure, success, failure, failure does not open the circuit, and three failures in a row do.
- [ ] A success while OPEN still closes the circuit only through the `min_healthy` rule from #1383.
- [ ] Single PR, <= 100 LOC.

## Severity

P2: the breaker's threshold does not mean what its configuration says, so any long-lived group eventually trips it. Blocked on #1383, which touches the same method.

mapyr:
## Why

Closes #1390.

A group's circuit breaker opened after `circuit_failure_threshold` failures (default 10) counted over the whole life of the process, not after that many in a row. `CircuitBreaker` resets its count on `record_success()` while CLOSED. `McpServerGroup` never called that on a closed circuit: `report_failure()` counted every failure, and only `rebalance()` reset. So any long-lived group eventually opened its circuit while its members were healthy.

## What

- `report_success()` calls `record_success()` when the circuit is not open, in the new `_end_failure_run()`. A success now ends the run of failures, as the breaker intends.
- An **open** circuit is left to `_maybe_close_circuit()` (#1383). Calling `record_success()` in OPEN would close the circuit at once and skip its `min_healthy` check.
- **HALF_OPEN** is handled on purpose: it takes the breaker's rule, and a success closes it.
  - The group never reaches HALF_OPEN by itself. The breaker enters it only in `allow_request()`, and the group does not call that.
  - The one way in is a snapshot restored by `_restore_group_circuit_breakers` that the group did not write. `save_group_circuit_breakers` persists only group breakers, which never half-open.
  - In HALF_OPEN the group already reports the circuit closed (`is_open` is False) and routes through it.
  - If a success left it alone, it would stay half-open for good, and one failure would open it.
  - A unit test pins each half: the group never half-opens by itself, and a restored half-open circuit closes on a success.

A success on **any** member ends the run. The circuit belongs to the group: it opens when the group fails in a row, and single members are handled by rotation. A member's start also counts, because the saga reports `McpServerStarted` as a success.

This changed one assumption in the #1383 harness. `pair` mode set `failure_threshold: 4` and relied on the count running across math-b's start. It now uses `2`, which is math-b's own run, so the circuit still opens once the last member is out. The run also changed: the circuit now opens when math-a goes out, math-b's start closes it again (min_healthy 1), and math-b's two failures open it once more. I confirmed this with a trace of the group's reports.

### Tests

- `tests/unit/test_a_group_circuit_counts_failures_in_a_row.py`:
  - With threshold 3, failure, failure, success, failure, failure leaves the circuit closed (count 2), and a third failure in a row opens it (`GroupCircuitOpened.failure_count == 3`).
  - Scattered failures never open it, however many there are.
  - A success on another member ends the run.
  - An OPEN circuit is not closed by a success below `min_healthy`, and closes at `min_healthy`.
  - The two HALF_OPEN tests.
- `tests/integration/test_scattered_health_check_failures_do_not_open_a_group_circuit.py` is a served-path test in #1383's style. It adds a `scattered` mode to `_group_recovery_harness.py`:
  - The real `bootstrap()` and its health-check worker, saga and group run it.
  - The health checks go fail, fail, pass, three times, then a third failure in a row.
  - `tests/mock_provider.py` fails `tools/list` while a flag file exists. The harness sets the flag for the next check as it hears each one, so the order does not depend on timing.
  - The run records the group after each check: the circuit stays closed through six scattered failures, the count goes `1,2,0,1,2,0,1,2`, and the ninth check opens it with a count of 3.
  - The `tools/list` branch moved into a helper, so `main()` stays at its baselined CC 16.

## How tested

- The new tests fail without the fix. I reverted `src/`, ran them, and re-applied:
  - Unit: 4 of 8 fail (the in-a-row case, both scattered cases, and the restored HALF_OPEN). The other 4 are regression guards and pass either way.
  - Served path: `test_scattered_failures_never_open_the_circuit` and `test_a_passing_check_starts_the_count_again` fail, with the count going `[1, 2, 2, 3, 3, 0, ...]`. The circuit opened at the fourth check, then the next pass closed it again under `min_healthy`.
- With the fix, in a private venv (`mcp_hangar.__file__` points into this worktree):
  - `ruff check src tests`: pass
  - `ruff format --check src tests`: pass
  - `mypy src/mcp_hangar`: no issues in 429 files
  - `lint-imports --config .importlinter`: 1 kept, 0 broken
  - `python scripts/check_dead_symbols.py`: baseline holds
  - `python scripts/build_changelog.py check`: OK
  - `pytest tests/integration`: 402 passed, 5 skipped
  - `pytest tests/unit`: 7905 passed, 4 skipped. `test_there_are_dockerfiles_to_check`, which usually fails under `.claude/worktrees` for environmental reasons, did not fail in this run.

## Risk and rollback

Low. One method in `McpServerGroup` now resets the circuit's failure count on a success while the circuit is not open. The visible change is that circuits open less often: only on `failure_threshold` failures in a row. Revert the single commit.

Not changed here, possible follow-ups:

- `circuit_reset_timeout_s` has no effect on a group. The group never calls `allow_request()`, so the timeout is never consulted.
- `_restore_group_circuit_breakers` loads `store.load("circuit_breaker")` without the group id, so every group gets the same restored row.

## Agent metadata

- [x] LOC declared upfront: `src/` +32/-2, under the 100 in the brief. Tests are separate.
- [x] Files in scope match the issue brief (`mcp_server_group.py`, `tests/unit/`), plus the served-path test the task asked for: the #1383 harness, its mock provider and a new integration test.


mapyr:
Reviewed and ready.

- **A success on a closed circuit resets the failure count.** `circuit_failure_threshold` now means failures in a row, as the breaker was written to count them. An open circuit still closes only through the `min_healthy` rule in `_maybe_close_circuit()` (#1383).
- **HALF_OPEN is handled deliberately.** The group never gets there by itself, because it does not call `allow_request()`. A restored half-open snapshot now closes on a success rather than staying half-open for good.
- **Test coverage.** #1383's `pair` harness still drives open, close and open again; only its threshold moved. The new served-path `scattered` mode shows six scattered health-check failures that do not open the circuit, and three in a row that do.
- **Worth knowing, but not introduced here.** A member's start counts as a success, the saga's semantics since #1383, so it also ends a failure run.

Pre-existing follow-ups are tracked in #1388 (restore without a group key) and #1398 (`circuit_reset_timeout_s` has no effect). CI is fully green.

Issue #1325, opened 2026-09-11
No metric separates a provider the gateway gave up on from one that is merely idle


mapyr:
## The gap

Ten minutes after a provider becomes unreachable, nothing in the metrics
distinguishes it from a provider that is simply idle. Both read
`mcp_hangar_mcp_server_state = 0` (cold), both read
`mcp_hangar_mcp_server_up = 0`, and neither has any counter moving. An alert
cannot tell "the gateway gave up on this server" from "nobody has called this
server lately", and in a lazy-start deployment the second is the normal state
of most of the pool.

## How it looks, measured

I cut one remote provider off from the gateway (a CiliumNetworkPolicy denying
ingress to its port) and watched, on a 2-replica environment with
`max_consecutive_failures: 3` and `health_check_interval_s: 30`:

```
t+0      provider unreachable
t+1m     mcp_hangar_health_check_consecutive_failures = 2   <- and it stays there
t+1.5m   state passes through 3 (degraded): MCPHangarProviderDegraded fires
t+2m     the gateway gives the server up; health checks STOP
t+2m..   health_checks_total stops incrementing for it
         consecutive_failures gauge frozen at 2, forever
         /api/mcp_servers/<id> reports state: cold, alive: false
t+10m    increase(...[10m]) windows have emptied.
         Metrically the provider is now indistinguishable from an idle one,
         while still being completely unreachable.
```

Two consequences, both of which cost me real debugging time:

1. **Thresholds above `max_consecutive_failures - 1` can never fire.** The
   consecutive-failure gauge is capped by the give-up rule, so the
   `> 5` / `> 2` style rules shipped in the chart's alerts are unreachable on
   any installation with `max_consecutive_failures: 3`. A rule that looks
   correct in the file never fires, and nothing says so.
2. **A persistent outage self-resolves.** Any rule written on
   `increase(health_checks_total{result="unhealthy"}[10m])` -- which is the
   only counter that moves during the failure -- goes quiet after the window
   rolls past the last probe, so the alert clears while the provider is still
   down.

## Why the existing signals do not close it

- `mcp_hangar_mcp_server_state` has a `dead` value (4, per the gauge's own
  description) but the given-up server reads 0. Whatever path takes it out
  returns it to cold.
- `mcp_hangar_mcp_server_stops_total` has a `reason` label, and the only reason
  this installation has ever observed is `idle`. Grepping the source, the call
[...]

PR #1399, opened 2026-09-13
feat(core): leave a given-up server dead and expose its last-healthy time

### LINKED ISSUE #1359: feat(core): expose a last-healthy timestamp that survives a server going cold
## Background

Split out of #1325. Measured on core 2.18.2 with remote providers, `max_consecutive_failures: 3`, `health_check_interval_s: 30`.

## Why

Ten minutes after a provider becomes unreachable, nothing distinguishes it from a provider nobody has called. Both read `mcp_hangar_mcp_server_state = 0` (cold), both read `mcp_hangar_mcp_server_up = 0`, and no counter is moving for either — because once the gateway gives a server up, health checks stop:

```
t+2m     the gateway gives the server up; health checks STOP
t+2m..   health_checks_total stops incrementing for it
         consecutive_failures gauge frozen at 2, forever
t+10m    increase(...[10m]) windows have emptied
```

In a lazy-start deployment "nobody called it lately" is the normal state of most of the pool, so alerting on the existing signals means alerting on normal operation. A persistent outage also self-resolves: any rule on `increase(health_checks_total{result="unhealthy"}[10m])` goes quiet once the window rolls past the last probe.

## What changes

A timestamp gauge that survives the server going cold, e.g. `mcp_hangar_mcp_server_last_healthy_timestamp_seconds`.

Then "down and nobody noticed" is `time() - last_healthy > 15m`, which stays true while the condition stays true, and an idle-but-fine server keeps a recent value because its last probe succeeded. This is the smallest of the three options in #1325 and the only one that needs no contract decision.

## Acceptance criteria

- [ ] The gauge is set on every successful health check and is not reset, cleared or removed when the server goes cold or is given up.
- [ ] The gauge is registered in `_register_all_metrics()` and a test asserts it appears on `/metrics` (see #1059).
- [ ] A test covers the give-up path: after the gateway stops probing, the gauge retains the last successful probe's timestamp.
- [ ] An idle server that has never failed reports a recent value, so `time() - last_healthy` separates idle from given-up.
- [ ] The intended PromQL is documented alongside the metric.

## Out of scope

Stop reasons (#1325 option 2) and the `dead` vs `cold` enum question (#1325 option 3) — both split out separately. Chart alert rules live in `mcp-hangar/helm-charts`.

## References

- https://github.com/mcp-hangar/mcp-hangar/issues/1325
- https://github.com/mcp-hangar/mcp-hangar/issues/1059

### LINKED ISSUE #1361: chore(core): decide whether a given-up server reads dead or cold
## Background

Split out of #1325. This is the contract question the other two children (#1359, #1360) deliberately avoid.

## Why

`mcp_hangar_mcp_server_state` has a `dead` value (4) in the gauge's own description, but a server the gateway has given up on reads 0 (cold). Whatever path takes it out returns it to cold, so the enum does not mean what it documents, and the only value that would have made the outage legible is never emitted.

Leaving the state at `dead` is the most direct fix. It is also a contract change: every alert rule, dashboard and client reading that gauge has been written against an enum where 4 never appears, and `cold` currently covers both "not started yet" and "given up on".

## What changes

Decide whether the gauge keeps its documented enum or the documentation is corrected to match behaviour, and whether a given-up server is a distinct state or a cold server with a separate signal (#1359, #1360) marking it.

Also worth resolving here: whether `dead` is terminal, and what returns a server from it — a successful call, an explicit start, or an operator action.

## Acceptance criteria

- [ ] The decision is recorded, including whether the gauge's enum changes or its description does.
- [ ] If the enum changes: the transition into and out of `dead` is defined and tested, and the change is named in the upgrade note as affecting anything that reads the state gauge.
- [ ] If the enum does not change: the gauge description is corrected so `dead` is not documented as reachable when it is not.
- [ ] Either way, chart alert expressions that read the state gauge are re-checked against the outcome.

## Out of scope

Implementing the timestamp gauge (#1359) and the stop reason (#1360); both are independent of this decision and can land first.

## References

- https://github.com/mcp-hangar/mcp-hangar/issues/1325

mapyr:
## Why

A server Hangar gave up on read `mcp_hangar_mcp_server_state = 0` (cold). That is the state of a server nobody has called, so an outage looked like an idle pool (#1325). On #1361 the maintainer decided to emit `dead`, and to land it together with #1359's last-healthy timestamp, so alert rules have an expression that survives the change.

Closes #1359
Closes #1361

Rebased on `main` after #1397 (a group circuit counts failures in a row) and #1402 (the `collect_events` drain fix). The drain fix is on `main` now and no longer part of this diff.

### How each failure path ends now

| Path | Before | Now |
| --- | --- | --- |
| Recovery saga, `max_retries_exceeded` | `StopMcpServerCommand` → `shutdown()` → `McpServerStopped("shutdown")` → `cold` (0) | `GiveUpOnMcpServerCommand` → `McpServer.give_up()` → `dead` (4), reason `given_up`, connection closed, tool catalogue cleared |
| A health check finds the process gone | `DEAD` by assignment with no event, so the gauge stayed `2` and `up` stayed `1` | `dead`, reason `crashed`, what is left of the connection closed |
| A start fails below `max_consecutive_failures` | `DEAD` by assignment with no event, so the gauge stayed `1` | `dead`, reason `start_failed` |
| Capability drift in block mode | `dead`, already with an event | `dead`, reason `capability_blocked` |
| Health checks fail past the threshold | `degraded` (3), and the saga retries | unchanged |
| A stop lands while a start is under way | the failed start made it `dead` | stays `cold` |
| `ready` read back from the event log, with no connection in the new process | `dead` on its first health check | `cold` |

The aggregate records the reason, and `McpServerStateChanged.dead_reason` persists it. The new field is optional, so a stream written before it existed still loads, with no reason. `McpServer.dead_reason_snapshot` reads the reason without the lock. `DEAD_NOT_ROUTED_BY_GROUPS` and `DEAD_NOT_REVIVED_BY_CALLS` name the two rules below.

### What starts a dead server again

| Why it died | A call through a group | A call naming it | A deliberate start |
| --- | --- | --- | --- |
| `given_up` | never | yes, once its backoff has passed | yes |
| `capability_blocked` | never | never | yes |
| `crashed`, `start_failed` | yes, once its backoff has passed | yes, once its backoff has passed | yes |

- **A call respects the server's backoff.** The executor judges a dead target by why it died and by `health.can_retry()`, not by its failure count. It does so at the circuit-breaker gate, and again at the cold-start gate, because the approval gate can hold a call in between. Inside the backoff the call is refused with `CircuitBreakerOpen` and the time to wait. `McpServer.ensure_ready(by_call=True)`, which `invoke_tool` uses, applies the same rules.
- **A refused group member counts as failed.** A failed start, a call refused inside the member's backoff, and a degraded member each count as that member's failure, as a failed invocation does. So a member whose restart keeps failing leaves rotation at `unhealthy_threshold`, and the group fails over.
- **Deliberate starts don't wait.** They are `hangar_start` on the server or on its group, `StartMcpServerCommand`, `POST /api/mcp_servers/{id}/start`, `hangar_warm` naming the server, a group adding the server with auto-start, and a failover saga starting the backup it was configured with.
- **Nothing else starts a dead server:**
  - The health worker skips it. `get_health_check_interval("dead")` is now `0.0`; it was 60.
  - The recovery saga cancels its pending restarts on a give-up, a start and a stop.
  - The front door's boot warm-up skips it, and so does `hangar_warm` with no names, which now returns `skipped_dead`.
  - `hangar_tools` lists a dead server, or a group's dead member, without starting it.
  - The GC acts only on servers that are `ready`.
- **Groups.** Selection skips members whose reason is in `DEAD_NOT_ROUTED_BY_GROUPS`. A crashed member is selected and restarted, as on `main`. The group saga handles `McpServerStateChanged` into `dead`: it takes a member it may not route to out of rotation, and a successful start puts it back, subject to `healthy_threshold`. `healthy_count` leaves every dead member out, so the group's own state stays true.

### The last-healthy timestamp

`mcp_hangar_mcp_server_last_healthy_timestamp_seconds{mcp_server}` is defined in `metrics.py` and registered in `_register_all_metrics()`.

- `MetricsEventHandler` writes it on the served path from `HealthCheckPassed`, `McpServerStarted` and `ToolInvocationCompleted`, at each event's `occurred_at`.
- It never moves backwards (`Gauge.set_max`), and going cold or dead doesn't clear it.

### Metric series

- **A removed server loses its gauges.** When a server is deleted, unloaded or reloaded away, its lifecycle gauges are removed (`Gauge.remove`, `remove_mcp_server_series`). Counters stay.
  - A deletion removes them on every replica: `remove_series_of_deregistered` is a projection on `McpServerDeregistered`.
  - An unload or a reload is local to its replica, so `MetricsEventHandler` removes those, on `McpServerHotUnloaded` and `ConfigurationReloaded.mcp_servers_removed`.
- **Delete stops a dead server too.** Before, it stopped only running ones.
- **Boot restores the gauges of failed servers.** After a restore from a durable event store, a server restored `dead` or `degraded` has its state and last-healthy time put back on its gauges. A restored `ready` one has no process in the new one, so it gets no series until its first check or call makes it `cold`.

### Upgrade note and changelog

The `## Next —` draft in `UPGRADE.md` covers:

- the `state`, `up` and `initialized` series, before and after, for each path;
- that an alert on `state == 0` misses dead servers, and what to use instead;
- that `sum(mcp_hangar_mcp_server_up) == 0` can newly fire;
- the revival table and the full list of deliberate starts.

The changelog entry is `changelog.d/1361-given-up-server-reads-dead.changed.md`.

## Changes since the second review

1. **A crashed member whose restart keeps failing now fails over.** A refused group member counts as failed, whether the refusal comes from the cold-start gate or is a `CircuitBreakerOpen`.
   - Tests: `TestAGroupMemberTheBatchRefuses`, and a new served-path mode, `failover`. It runs two members with every group threshold at its default. B's process is killed while its upstream is broken, then 20 calls go through the group. B fails exactly twice (`McpServerStartError`, then `CircuitBreakerOpen`) and leaves rotation, and the remaining calls go to A.
   - I didn't add a failover inside the same call. The refused call still fails. That is what `main` did, and re-selecting mid-pipeline would mean re-running the policy gates for the new member.
2. **A capability block is left to deliberate starts.** A group never routes to it and a call to it is refused. Tests: `TestACapabilityBlock`.
3. **`hangar_tools` doesn't start a dead server or dead member.** Tests: `TestHangarTools`.
4. **Boot seeds only `dead` and `degraded`.** Test: `test_a_restored_ready_server_is_not_put_on_its_gauges_as_up`.
5. **A server deleted on another replica loses its gauges here.** Tests: `test_a_server_deleted_on_another_replica_loses_its_gauges_here` and `test_bootstrap_subscribes_the_removal_as_a_projection`.
6. **Rebased after #1397.** `report_success` keeps #1397's `if is_open: _maybe_close_circuit() / else: _end_failure_run()`, and `_update_state()` runs after both branches. Groups now test membership in `DEAD_NOT_ROUTED_BY_GROUPS`, which a bare `MagicMock` attribute never satisfies. So no mock needs a `given_up_snapshot`, and #1397's tests, including the half-open one, pass unchanged.
7. **The mutants that survived are killed:**
   - M13: `test_hangar_warm_of_everything_skips_a_dead_server_and_naming_it_starts_it`.
   - M19: the crashed-member test asserts the group reads `healthy` after the restart.
   - M22: `test_deleting_a_dead_server_closes_a_connection_it_still_holds`.

**Nits.** A late event can write a series again after it was removed; that's documented on `remove_mcp_server_series`. The cold-start gate re-checks the backoff after the approval gate, as covered in item 1.

## Deviation from #1359's acceptance criteria

#1359 says an idle server that never failed reports a recent value. That holds only while the idle server is still `ready` and being probed. A server reaped for being idle is `cold`, nothing probes it, and its value ages. So the documented rule is:

```promql
time() - mcp_hangar_mcp_server_last_healthy_timestamp_seconds > 900
  unless mcp_hangar_mcp_server_state == 0
```

It uses default matching. `on(mcp_server)` would drop `instance`, so with several replicas a server that is cold on one replica would hide it being dead on another. A server that was never healthy has no series; `state == 4` covers it.

## Follow-up: helm-charts

This PR doesn't edit `mcp-hangar/helm-charts`. The only rule there that reads `mcp_hangar_mcp_server_state` is in `mcp-hangar/files/prometheus-alerts.yaml`. No recording rule reads it, and neither does the operator chart. No chart rule or dashboard reads `mcp_hangar_mcp_server_initialized`.

| Rule | Expression today | Needs |
| --- | --- | --- |
| `MCPHangarProviderDegraded` | `mcp_hangar_mcp_server_state == 3`, `for: 5m` | Still right for degraded. A server the saga gives up on now moves from 3 to 4, so this alert resolves as the server dies. Keep it, and add the next rule. |
| new: `MCPHangarProviderDead` | none | `mcp_hangar_mcp_server_state == 4`, `for: 1m`, `severity: critical` |
| new: `MCPHangarProviderNotSeenHealthy` | none | `time() - mcp_hangar_mcp_server_last_healthy_timestamp_seconds > 900 unless mcp_hangar_mcp_server_state == 0`, `for: 5m` |
| `MCPHangarAllProvidersDown` | `sum(mcp_hangar_mcp_server_up) == 0 and sum(mcp_hangar_mcp_server_info) > 0` | **Can newly fire.** A crashed server used to keep reading `up` 1, and now reads 0. The rule already fires for a pool that is all cold (#1325). To fire only when something failed: `sum(mcp_hangar_mcp_server_up) == 0 and count(mcp_hangar_mcp_server_state == 4) > 0`. |

Two related rules don't read the state gauge: `MCPHangarProviderUnhealthy` (`consecutive_failures > 5`) and `MCPHangarHighConsecutiveFailures` (`> 2`). Neither can fire above `max_consecutive_failures - 1` (#1325). `MCPHangarProviderDead` is what they were trying to catch.

The dashboards need no change. The overview "Providers" table and the provider-details "State" stat already map `4` to `DEAD`, and the state-history panels plot raw values.

## Found, not fixed

With default settings, the recovery saga's own restarts are refused by the server's backoff, and nothing reschedules them: #1401. The served-path harness stays clear of it in its `single` and `group` modes.

## How tested

- `tests/unit/test_a_given_up_server_reads_dead.py` and `tests/unit/test_what_revives_a_dead_server.py`: the transitions and the reasons, the backoff for calls, the capability block, the saga, groups and the group saga, refused group members, series removal including a deletion tailed from a peer, boot seeding, the warm-ups, `hangar_tools`, the serializer round trip, and the nits.
- `tests/integration/test_a_given_up_server_reads_dead_on_the_served_path.py` (23 tests, about 13s). It runs the real `bootstrap()`, the served MCP app, and the real workers and sagas, with real stdio upstreams, in three modes: `single`, `group`, and `failover` at default thresholds. It passed in 3 of 3 back-to-back runs, and inside the full integration run.
  - It reads what a command drove only once that command's own events have reached a watcher subscribed after the metrics handler and the sagas. The health and GC workers publish whatever a server has pending, so a command can return before its events are delivered. `decision-coverage` caught the harness reading the gauge after `hangar_stop` too early, and `227ca787` fixes that.
- `MCP_HANGAR_LIVE_VERIFY=1 pytest tests/live/test_t0_last_healthy.py tests/live/test_t0_smoke.py`: 6 passed, against a real `serve --http`.
- Gates on this branch, each exit code captured directly:
  - `ruff check src tests`, `ruff format --check src tests`, `mypy src/mcp_hangar`, `lint-imports --config .importlinter` and `scripts/check_dead_symbols.py` all exit 0.
  - markdownlint on `UPGRADE.md`: 0 issues.
  - `scripts/build_changelog.py check`: OK.
  - `promote_upgrade_notes.py promote`, run on a copy of `UPGRADE.md`, folds the draft.
- Full suites:
  - `pytest --ignore=tests/integration`: 8,280 passed, 65 skipped and 1 failed, exit 1. The failure is `test_there_are_dockerfiles_to_check` (`assert 0 >= 4`), which always fails under `.claude/worktrees`.
  - `pytest tests/integration/ --timeout=60`: 437 passed, 5 skipped, exit 0.

## Risk and rollback

This changes the contract of the state gauge and of group selection, and the upgrade note documents both. To roll back, revert the commits. `McpServerStateChanged` gains an optional field, and older streams still load.

## Agent metadata

- [ ] Single goal: two issues by design (#1361 lands with #1359), plus two reviews' findings
- [x] LOC declared upfront: about 1,300 lines changed in `src/` and tests combined across the three commits


mapyr:
This stays a draft while an independent review runs. Two things are also moving out of it before it is marked ready: the `AggregateRoot.collect_events()` drain-race fix gets its own issue and PR, because it affects every aggregate, and the three found-but-not-fixed items get their own issues.

mapyr:
An independent review of this PR is back. Keep it a draft; it has two blockers.

**Blockers**

1. **A call to a `dead` server starts it with no backoff.** The executor's gates let a dead target through on every call. With the upstream still broken, 50 rapid calls launched it 50 times in about 40 ms. On `main`, the same server ended `degraded`, and its backoff refused those calls. A call should reach a dead server only once its backoff has elapsed. Explicit starts stay exempt.
2. **A group member whose process crashed is never restarted by a call.** The decision on #1361 covers servers Hangar *gives up on*. This PR also sends a crashed process to `dead`, and group selection now skips dead members. On `main`, the next group call restarted such a member; here, a single-member group refuses every call until an operator acts. Given-up and crashed must be told apart, and a dead member has to leave rotation so `healthy_count` stays true.

**Should-fix**

- **Stale series after removal.** A deleted, unloaded or reload-removed dead server keeps `state == 4` and an ageing last-healthy series. Remove its series when the server is deregistered.
- **Restored dead servers at boot.** Nothing publishes gauges for a restored `dead` server, and the front-door boot warm-up revives it.
- **Multi-replica PromQL.** `unless on(mcp_server) … == 0` suppresses the alert whenever the server is cold on any replica. Use default matching, or `on(mcp_server, instance)`.
- **Incomplete upgrade note.** It must also name `mcp_hangar_mcp_server_up` and `mcp_hangar_mcp_server_initialized`, which change too.
- **"Nothing else revives it" is not true.** `hangar_warm()`, the failover saga, a group's `start_all` and `add_member`, and stale recovery timers all start a dead server. The stale timers need a fix in code: `_handle_started` forgets pending restarts without cancelling them.

**Verified OK**

- **Transitions:** `DEGRADED → DEAD`, and `give_up()` acts only from `degraded`.
- **Background workers:** the health worker skips dead servers without a busy loop, and GC and idle reaping act only on `ready`.
- **Timestamp:** it is written on the served path from three events and never moves back.
- **Mutation tests:** seven were run, and all seven were caught.

Fixes are in progress on this branch.

mapyr:
Two independent review rounds are resolved. I've reviewed this and am marking it ready.

**Round 1:**
- A call to a `dead` server now waits out its backoff; explicit starts are exempt.
- Crashed and given-up servers are told apart by `dead_reason`. Groups skip given-up members, and a crashed member stays restartable by a group call.
- Series of removed servers are dropped.
- Boot handling is fixed.
- The PromQL uses default matching.
- The upgrade note covers the `up` and `initialized` series.
- The overclaims are corrected, and stale recovery timers are cancelled.

**Round 2:**
- A refusal of a group member counts as that member's failure, so a member whose restart keeps failing leaves rotation and the group fails over. A served-path test covers this with default thresholds.
- A capability block is left to deliberate starts.
- `hangar_tools` no longer starts a dead server.
- Boot seeds only `dead` and `degraded` servers.
- A delete on another replica now drops this replica's series, through a projection handler.
- Rebased onto #1397 and #1402.
- Three test gaps found by mutation testing are closed.

**Harness fix (`227ca787`):** the `decision-coverage` failure was a race in the test harness. `hangar_stop` can return before its event reaches the metrics handler, so the harness now waits for the events a command caused. Product code is unchanged.

CI is fully green. Merging this closes #1359 and #1361. The upgrade note names the state-gauge change and the replacement alert rules. The helm-charts follow-up is listed in the body.

PR #1402, opened 2026-09-13
fix(core): drain aggregate events one at a time so none is lost

### LINKED ISSUE #1400: fix(core): collect_events() drops an event another thread records while it drains
## Symptom

`AggregateRoot.collect_events()` drops a domain event if another thread records it while the drain is running. The lost event is never published, so no subscriber sees it: not the metrics, not the sagas, not the audit handlers, and not the event store. Nothing logs an error.

Measured on `origin/main` (`f688e427`) with the real `AggregateRoot`. One thread records 20,000 events while another thread drains in a loop, with `sys.setswitchinterval(1e-6)` set to interleave them finely:

| Run | Recorded | Collected | Lost |
| --- | --- | --- | --- |
| 1 | 20,000 | 17,645 | 2,355 |
| 2 | 20,000 | 18,078 | 1,922 |
| 3 | 20,000 | 18,203 | 1,797 |

The same stress run against the same logic on another branch lost between 3,597 and 6,383 events per run.

This is reachable on the served path. The health worker drains a server's events on its own loop, while a recovery saga's scheduled restart records events on a timer thread. In the served-path test for #1361, one restart's `McpServerDegraded` was lost in exactly this way. The saga counts those events, so it never gave the server up. Every aggregate and every event type is exposed; which event is lost depends only on timing.

## Root cause

```python
# src/mcp_hangar/domain/model/aggregate.py, origin/main f688e427
    def _record_event(self, event: DomainEvent) -> None:
        ...
        self._uncommitted_events.append(event)

    def collect_events(self) -> list[DomainEvent]:
        ...
        events = list(self._uncommitted_events)
        self._uncommitted_events.clear()
        return events
```

An event appended between the copy and the clear is cleared without ever being returned. `collect_events()` is called from the background workers and the command handlers without the aggregate's lock.

## What

Drain the list with `list.pop(0)` until it is empty, instead of copying it and then clearing it. Each `pop` and each `append` is atomic, and appends only go on the end. So an event recorded during a drain is either returned by this call or left for the next one. This needs no new lock. A bare lock on `AggregateRoot` would sit outside the `TrackedLock` hierarchy, which `tests/unit/test_aggregates_are_in_the_lock_hierarchy.py` exists to prevent.

## Files in scope

```
src/mcp_hangar/domain/model/aggregate.py
tests/unit/test_an_event_recorded_while_draining_is_not_lost.py
changelog.d/<issue>-aggregate-drain-race.fixed.md
```

## Acceptance criteria

- [ ] A test records 20,000 events on one thread while another thread drains, and collects every event exactly once, in order.
- [ ] On a single thread, `collect_events()` returns the same events in the same order as before.
- [ ] No lock is added outside the `TrackedLock` hierarchy.
- [ ] Single PR, <= 100 LOC.

## Severity

P1 -- domain events are silently lost on the served path. Whatever a lost event drives (metrics, sagas, audit records, the event store) simply never happens.

Found while implementing #1359/#1361 (PR #1399).

mapyr:
## Why

`AggregateRoot.collect_events()` could drop a domain event that another thread recorded while it was draining. The dropped event was never published, so nothing downstream saw it. That covers the metrics, the sagas, the audit records and the event store, and no error was logged. On `origin/main`, recording 20,000 events on one thread while another drains loses 1,797 to 2,355 of them per run.

This was found while implementing #1359/#1361 (#1399). There the health worker drained a server's events while a recovery saga's restart recorded one on a timer thread. A `McpServerDegraded` was lost, and the saga never gave the server up. The fix is split out from #1399 because it affects every aggregate, and it needs its own changelog entry and its own revert path.

Closes #1400

## What

- `collect_events()` now drains the list with `list.pop(0)` instead of copying it and then clearing it. Each `pop` and each `append` is atomic, and appends only go on the end. So an event recorded mid-drain is returned either by this call or by the next one.
- No lock is added. A bare lock on `AggregateRoot` would sit outside the `TrackedLock` hierarchy, which `test_aggregates_are_in_the_lock_hierarchy.py` exists to prevent.
- On a single thread, it returns the same events in the same order as before. The only other reader of the pending list, `has_uncommitted_events()`, is unaffected.
- `tests/unit/test_an_event_recorded_while_draining_is_not_lost.py` records 20,000 events on one thread while another drains, with `sys.setswitchinterval(1e-6)`, and asserts every event is collected exactly once, in order. It also covers the single-thread case.
- `changelog.d/1400-aggregate-drain-race.fixed.md`.

## How tested

All gates ran on this branch (one commit on `origin/main` `f688e427`), each with its exit code captured directly:

- `ruff check src tests`, `ruff format --check src tests`, `mypy src/mcp_hangar`, `lint-imports --config .importlinter` and `scripts/check_dead_symbols.py` all exit 0.
- `pytest --ignore=tests/integration`: 8,050 passed, 63 skipped and 1 failed, exit 1. The failure is `tests/ci/test_base_images_are_pinned_and_updated.py::test_there_are_dockerfiles_to_check` (`assert 0 >= 4`). It always fails in a checkout under `.claude/worktrees`, where the Dockerfile glob finds nothing, and is unrelated to this change.
- `pytest tests/integration/ --timeout=60`: 401 passed, 5 skipped, exit 0.
- Before the fix, the same stress, run against `origin/main`'s real `AggregateRoot`, lost 2,355, 1,922 and 1,797 of 20,000 events in three runs.

## Risk and rollback

Low. The change is one method, and it touches every aggregate: every caller now gets every event, where before some were dropped under contention. `pop(0)` is O(n) per event, but the pending lists hold a handful of events per operation. To roll back, revert the single commit.

## Agent metadata

- [x] LOC declared upfront: about 20 lines in `src/`, a 63-line test, and a 9-line changelog fragment
- [x] Files in scope match the issue brief (#1400)


mapyr:
Reviewed and ready.

- `collect_events` used to copy the pending list and then clear it, so an event recorded on another thread between the two steps was cleared without ever being returned. It now drains with `list.pop(0)`. Each `pop` and `append` is atomic, and appends only ever go on the end, so an event recorded mid-drain is returned by this call or the next.
- There is no lock, which is deliberate: a bare lock here would sit outside the `TrackedLock` hierarchy.
- The stress test fails on the old copy-then-clear and passes on the drain.
- Scope is exactly three files: the change, its test and a changelog fragment. #1399 carries the same change and drops it on rebase.

One nit, not blocking: `pop(0)` is O(n) per event, so a drain is O(n²). That is fine for the handful of events an aggregate holds between drains; a `deque` would make it linear if that ever changes.

CI is fully green.

PR #1403, opened 2026-09-13
fix(core): govern a task a front-door flat call returns, as hangar_call does

### LINKED ISSUE #1394: fix(core): a task returned through a front-door flat call cannot be polled
## Symptom

On a front door, when an upstream answers a flat `tools/call` with a task handle, the client gets a handle it can never use. A later `tasks/get`, `tasks/cancel` or `tasks/update` on it answers "Task not found".

It fails closed: nothing leaks and no task is acted on for the wrong caller. But a long-running upstream tool is effectively unusable through the flat surface.

## Root cause

Only the `hangar_call` path registers relayed task handles with the governed task store. `_govern_relayed_tasks(executed)` (`server/tools/batch/__init__.py:192`) is called once, at `:597`, after `hangar_call` runs its batch.

The flat path, `_flat_call_tool` in `fastmcp_server/flat_tool_projection.py`, dispatches through the same `BatchExecutor` and returns `result.result` as is. Nothing records ownership of a task the upstream created, so the store never learns about it and `GovernedTaskStore.authorize` finds nothing to authorize.

This was found while building the security-predicate reachability test (#1386). That test drives `tasks/*` only after a `hangar_call`, which is why it passes.

## What

- Govern relayed task handles on the flat path the same way `hangar_call` does: record the owner, with tenant and principal, of any task handle in the result before returning it.
- Keep the ownership rules identical to the `hangar_call` path. The `tasks/*` handlers should not care which path created the task.

## Files in scope

- `src/mcp_hangar/fastmcp_server/flat_tool_projection.py`
- `src/mcp_hangar/server/tools/batch/__init__.py`, if `_govern_relayed_tasks` should move somewhere both paths can reach.
- `tests/`

## Acceptance criteria

- [ ] A front-door flat `tools/call` whose upstream returns a task handle can be followed by `tasks/get` from the same caller, which succeeds. `tasks/cancel` and `tasks/update` follow the same ownership rules as on the `hangar_call` path.
- [ ] Another tenant's `tasks/get` on that handle is refused exactly as on the `hangar_call` path.
- [ ] The test drives the served front door, not a hand-built executor.
- [ ] Once this lands, #1386's reachability registry declares the flat path for `GovernedTaskStore.authorize`.

## Severity

P2: fails closed, but tasks do not work through the front door's default surface.

mapyr:
## Why

Closes #1394. On a front door, when an upstream answered a flat `tools/call` with a task handle, nobody could use the task afterwards. Only `hangar_call` ran the seam that records a relayed task's owner, so `tasks/get`, `tasks/cancel` and `tasks/update` answered "Task not found" to every caller, the task's creator included. With the SDK Hangar ships (`mcp==2.0.0`), the call failed before that. The upstream's nested `{"task": {...}}` is not a `CallToolResult`, so the handler's wrap raised and the client got `-32602`, while the `front_door_tool_call` line logged `outcome=ok`.

## What

- **One seam, two paths.** `_govern_relayed_tasks` moves verbatim from `server/tools/batch/__init__.py` to `server/tools/batch/relay_seam.py` as `govern_relayed_tasks`. `hangar_call` calls it as before, and the flat call now calls the same function. The owner is the one `hangar_call` records: the captured tenant and principal. A task that cannot be governed gets the same failure on both paths (`TaskRelayNotSupported`, `TaskRelayRegistrationFailed`).
- **The flat decision gets its own module**, `fastmcp_server/flat_call_tasks.py`, because `register_flat_tool_handlers` sits at the C901 ceiling (15). The handler change is two lines and adds no branch.
- **The caller gets a task it can read.** A governed task is answered with the SEP-2663 `CreateTaskResult` (flat, `resultType: "task"`). `_call_v2_inner` passes it through instead of forcing it into a `CallToolResult`. Its fields come from the store's own parse (`mint_from_upstream`, now a `staticmethod`) and from the projection `tasks/get` uses, which moves out of `_flat_snapshot` into `task_wire_fields`. So the task id the caller gets is the one the ledger is keyed on.
- **The decision reads the worker's capture, never the upstream's shape.** Only a task the store recorded becomes a task result.
- **Reachability registry (#1386/#1393).** `_security_predicates.py` declares `front door tasks/get`, `tasks/cancel` and `tasks/update` for `GovernedTaskStore.authorize` and the session-suspension checks. The reachability test now mints its front-door task through the flat call and probes all three.

## How tested

- **New: `tests/unit/test_a_task_a_flat_call_created_can_be_polled.py`.** It drives the served app through the reachability harness, over streamable HTTP with real API keys; no executor is hand-built. Each topology creates the task its own way: `hangar_call` on egress, the flat call on the front door. Then the owner, a peer in the owner's tenant, and a caller in another tenant each send `tasks/get`, `tasks/cancel` and `tasks/update`. It asserts that:
  - the flat call answers with a `resultType: "task"` result whose fields match what `tasks/get` serves;
  - the owner is served on the front door;
  - the peer and the other tenant get `-32602 "Task not found"`;
  - for every method and every caller, the front door's answer (status and payload) equals the `hangar_call` path's.

  Five hand-built cases pin the refusals the served app cannot reach unpatched: a store that fails to register, no store, a plain result, and a task-shaped result nobody captured.
- **Negative check.** With `main`'s `flat_tool_projection.py` swapped back in, the new test and the reachability test fail: the front-door mint gets `-32602`.
- **Gates run locally, on Python 3.11:**
  - `ruff check src/mcp_hangar` (including C901), `ruff format --check`, `mypy src/mcp_hangar`, `lint-imports --config .importlinter` and `python scripts/check_dead_symbols.py` all pass.
  - `pytest --ignore=tests/integration`: 8251 passed, 64 skipped, 1 failed. The failure is `test_there_are_dockerfiles_to_check`, which always fails when run under a `.claude/worktrees` checkout; it is environmental and unrelated.
  - `pytest tests/integration/ --timeout=60`: 414 passed, 5 skipped.
  - decision coverage: `pytest tests/unit tests/integration -m "not benchmark and not live" --cov=mcp_hangar --cov-report=json:coverage.json` gives 8541 passed and 9 skipped, and `python scripts/check_decision_coverage.py coverage.json` reports that all 31 decision-path modules hold their floor.
  - `scripts/check_changelog.sh` passes on the one fragment.

## Risk and rollback

Low risk. The `hangar_call` path runs the moved seam unchanged, and its tests only follow the new module path. On the flat path, a result without a captured task is handled exactly as before. Revert the single squash commit.

## Agent metadata

- [x] LOC declared upfront: +587/-190, of which about 150 is the verbatim seam move and 235 the new test
- [x] Files in scope match the issue brief (plus the `tasks/get` projection helper, shared rather than duplicated)


mapyr:

- **Shared seam.** A flat call now runs the same code `hangar_call` runs to register the task owner: `govern_relayed_tasks`, moved without changes to `relay_seam.py`. It runs on the main loop, as ADR-014 D4 requires. The owner's `tasks/get`, `tasks/cancel` and `tasks/update` are served. A peer in the same tenant and another tenant are refused, exactly as on the `hangar_call` path, and the test compares both paths.
- **Task results only for recorded tasks.** Whether the caller gets a task result depends on what the worker captured and the seam governed, never on the shape of the upstream's reply. An upstream cannot hand the caller a task the store has not recorded.
- **`-32602` on SDK 2.0.0.** The upstream's nested handle is not a `CallToolResult`, so flat calls that got a task back failed with `-32602`. They now return the SEP-2663 task result, built by the same parse that registered the task.
- **Coverage.** The security-predicate registry (#1393) declares the front-door `tasks/*` paths for `GovernedTaskStore.authorize`, and the reachability test probes them.

Follow-ups filed: #1404 (the call log records `ok` for a result the SDK later rejects) and #1405 (the current flat task shape, and task results sent to callers that did not negotiate tasks). CI is fully green.

PR #1406, opened 2026-09-13
feat(core): measure projected surface size, composition and churn

### LINKED ISSUE #1369: feat(core): measure projected surface size, composition and churn
## Background

Part of #1365.

`mcp_hangar_projected_tools` was introduced in #912 as the first measurement of what the governance surface costs in a client's context budget. Nothing today distinguishes the parts of that surface from each other, and nothing measures how often a projection changes — which is the whole subject of this epic.

## What changes

Extend the projected-surface measurement so a change in projection is observable, and the composition of the surface can be broken down.

## Read this before implementing

#1059: five metrics were defined, incremented, documented with ready-made PromQL in the project's own observability guide, and never added to `REGISTRY`. Every query in that guide returned empty on every deployment for seven versions. `PROJECTED_TOOLS` — the one this issue extends — was among them.

This is the third metric issue filed this month carrying the same warning. The acceptance criterion below is the point of the issue, not a formality.

2.19.0 also removed four metrics that nothing had written to since #1002, so the opposite failure has precedent too: register it and write to it, or do neither.

## Acceptance criteria

- [ ] A test asserts the metric appears on a real `/metrics` response, not that `_register_all_metrics()` was called.
- [ ] A test asserts it is incremented or set on the real serving path, not only in a unit fixture.
- [ ] Projection changes are countable, so "did the fleet change during this session" is answerable from Prometheus.
- [ ] Cardinality is bounded and carries no tenant label, consistent with #895's reasoning.
- [ ] Whatever PromQL is documented is executed in a test against a scrape of a running gateway, so the guide cannot drift from the exposition again.

## Out of scope

Trace exemplars (#1290, #1302). Per-call logging on the projected surface — that is #1362, and this issue depends on it for correlating a count with the calls that caused it.

## References

- #1059, #912, #895, #1362

mapyr:
## Why

`mcp_hangar_projected_tools` counted the tools a front-door `tools/list` handed a client, split by kind. It could not answer three things:

- How many bytes that is. The epic's diagnosis was "46 tools and 174 KB".
- Which upstream the bytes come from.
- Whether the list a caller holds is still the list it would be served now.

That last question is the subject of the epic. This PR makes all three answerable from Prometheus, and every number is read back from a real `/metrics` scrape. #1059 found `PROJECTED_TOOLS` among five metrics that were defined, incremented and documented but never registered.

Closes #1369
Part of #1365

## What

- **`mcp_hangar_projected_surface_bytes{kind}`** is a histogram of the bytes of tool definitions one listing hands the client.
  - `kind` is `governed` or `management`, the same split as `mcp_hangar_projected_tools`.
  - It measures each definition as compact UTF-8 JSON, dumped as the modern transport dumps it (`by_alias`, `exclude_none`).
  - The integration test asserts that the measured bytes equal the bytes the client actually received.
- **`mcp_hangar_projected_upstream_bytes{mcp_server}`** is a histogram of one upstream's share of a listing that includes it. This is the composition.
  - A group member is counted under its group id, so a listing served from either member lands on one series.
- **`mcp_hangar_projection_changes_total`** is a counter with no labels. It counts listings that served an identity a projection different from the one this replica last served that identity.
  - "Different" is `Projection` equality from #1381, compared as a digest that keeps that equality. It is order-independent: same names, same routes, same definitions.
  - The identity is the one #1395 keys on: tenant plus principal, narrowed by session.
  - An empty listing is remembered too, so going from nothing to 46 tools counts as a change.
  - A front door seeds the counter at `0` when it installs its handlers. A series that first appears at 1 reads as no increase to `increase()`.
  - The per-identity memory is an LRU of at most 1024 digests of 32 bytes each, the same bound as `served_tool_names`.
- All three are written from one call, `projection_metrics.observe_served_listing`. It runs only in the real-listing branch of `_list_projected_tools`, never on the SDK's pre-dispatch `tools/list` on a `tools/call` (#1049). `PROJECTED_TOOLS` moved into it too, so the function at the complexity ceiling got shorter.
- `lifecycle.metrics_endpoint` moved to module level, unchanged, so a test scrapes through the endpoint `serve --http` mounts rather than the registry.
- The served front-door harness from the #1368 test was extracted to `tests/integration/_front_door_harness.py`, unchanged apart from gaining `GET /metrics`, and is now shared by both tests.

### Cardinality

- None of the three carries a tenant label, for #895's reason: a public front door has unbounded tenant cardinality.
- #895 does not argue against per-server labels. `mcp_server` is the configured fleet, as on `mcp_hangar_tool_calls_total` and every other per-server metric.
- `kind` has two values.
- The integration and live tests assert that no series of these families carries any label but `kind`, `mcp_server` and `le`, and that no line names a tenant.

### What the counter cannot see, by design

- An identity's first listing on a replica: there is nothing to compare it with.
- A caller without a tenant, which is served nothing in front_door.
- An identity evicted from the LRU, whose next listing counts as a first listing.
- A change seen only by landing on another replica. The memory is per replica, like #877.

It can miss a change. It cannot count one that did not happen.

## How tested

The PromQL the guide will print lives in `tests/_promql.py::PROJECTION_QUERIES` and is **executed** against two `/metrics` scrapes of a running gateway. There is no Prometheus in CI, so `tests/_promql.py` is a small strict evaluator. It parses scrapes with `prometheus_client`'s own parser. It covers selectors, `rate`, `increase`, `sum`/`max`/`min` with `by`, `histogram_quantile` using Prometheus's interpolation, arithmetic, and scalar filters. It refuses anything else. A selector that matches no series is an error, not an empty result, which is exactly the drift #1059 found. The evaluator has its own tests for both halves: it computes what Prometheus computes, and it refuses what it cannot run faithfully.

Tests:

- `tests/integration/test_the_projected_surface_is_measured.py` runs over the real streamable-HTTP transport and the served composition, and every number comes from `GET /metrics`:
  - the families are typed and sampled on the scrape;
  - the bytes measured equal the bytes on the wire;
  - a policy edit under a connected client counts exactly once, and unchanged re-listings count zero;
  - a withdrawal for tenant A counts once, and tenant B's unchanged listing adds nothing;
  - the SDK's pre-dispatch listing moves none of the families and no memory;
  - no tenant label appears on any series;
  - every documented query answers from two scrapes, including `changed` answering "yes" and then "no".
- `tests/live/test_t0_projection_metrics.py` runs against a real `mcp-hangar serve --http` in front_door. The projection is changed through the REST withdraw endpoint. The test asserts +1 change, 3 listings, and the per-upstream series, and runs every documented query against the process's own `/metrics`.
- `tests/unit/test_the_projected_surface_is_measured.py` covers the parts: digest equality matches `Projection` equality, the LRU is bounded, the empty-to-full listing counts, eviction never invents a change, and a group is counted under its group id.
- `tests/unit/test_the_promql_evaluator_is_strict.py` covers the evaluator.

Commands run locally:

```
ruff check src/mcp_hangar                    # All checks passed
ruff format --check src/mcp_hangar           # already formatted
mypy src/mcp_hangar                          # Success: no issues found in 433 source files
lint-imports --config .importlinter          # 1 kept, 0 broken
python scripts/check_dead_symbols.py         # baseline holds: 12 unreferenced, 25 exported-unreferenced
ruff check src --select C901                 # no new findings; complexity baseline test passes
make changelog-check                         # OK
pytest tests/unit/test_every_metric_is_registered.py tests/unit/test_metric_headers_name_the_sampled_family.py \
  tests/unit/test_metrics_exposition.py tests/unit/test_complexity_baseline.py tests/unit/test_import_contracts.py
MCP_HANGAR_LIVE_VERIFY=1 pytest tests/live/test_t0_projection_metrics.py tests/live/test_t0_front_door_call_log.py \
  -m "live and t0" -o addopts=""             # 2 passed
pytest --ignore=tests/integration             # 8254 passed, 65 skipped, 1 failed (see below)
pytest tests/integration/ --timeout=60        # 421 passed, 5 skipped
pytest tests/unit tests/integration -m "not benchmark and not live" --cov=mcp_hangar ...
                                              # 8551 passed, 9 skipped (decision-coverage selection, one session)
python scripts/check_decision_coverage.py coverage.json   # all 31 decision-path modules hold their floor
```

The single failure in the `test` selection is `tests/ci/test_base_images_are_pinned_and_updated.py::test_there_are_dockerfiles_to_check`. It is environmental: it always fails when run from a `.claude/worktrees` checkout, and it is unrelated to this change.

The first decision-coverage run caught a real test bug. In one session with the unit tests, the process holds `mcp_hangar_projected_upstream_bytes` series for upstreams that no listing in the window included, and those read `0/0 = NaN`, exactly as an idle upstream does in Prometheus. The assertion now checks the `store` series and requires the others to be NaN. It was fixed and re-run green.

## Docs follow-up

This belongs in `mcp-hangar/docs` `guides/OBSERVABILITY.md`, as a new `#### Front Door Projection` subsection under "Currently Exported Metrics" after "Approval Gate". The queries must be copied **verbatim**: they are the strings in `tests/_promql.py::PROJECTION_QUERIES`, and the tests in this PR run them against a scrape of a running gateway. Changing one without the other is how #1059 happened.

````markdown
#### Front Door Projection

*`mcp_hangar_projected_surface_bytes`, `mcp_hangar_projected_upstream_bytes` and
`mcp_hangar_projection_changes_total` since <next version>.*

What a `front_door` gateway hands a client in `tools/list`, which sits in the
client's context on every turn. Only listings the client received are measured:
the SDK's own listing before a `tools/call` is not. None of these carries a tenant
label: a public front door has unbounded tenant cardinality.

| Metric | Type | Labels | Description |
| -------- | ------ | -------- | ------------- |
| `mcp_hangar_projected_tools` | Histogram | kind | Tools per listing: `governed` (upstream) or `management` (`hangar_*`) |
| `mcp_hangar_projected_surface_bytes` | Histogram | kind | Bytes of tool definitions per listing, as compact JSON |
| `mcp_hangar_projected_upstream_bytes` | Histogram | mcp_server | Bytes one upstream adds to a listing that includes it; a group reads as its group id |
| `mcp_hangar_projection_changes_total` | Counter | - | Listings whose projection differed from the one the same caller was last served on this replica |

**Example queries:**

```promql
# Did the projection served to any caller change on this replica in the last hour?
increase(mcp_hangar_projection_changes_total[1h]) > 0

# Tools a client is handed per listing, by kind
sum(rate(mcp_hangar_projected_tools_sum[5m])) by (kind) / sum(rate(mcp_hangar_projected_tools_count[5m])) by (kind)

# Bytes of tool definitions a client is handed per listing, by kind
sum(rate(mcp_hangar_projected_surface_bytes_sum[5m])) by (kind) / sum(rate(mcp_hangar_projected_surface_bytes_count[5m])) by (kind)

# 95th percentile of the governed surface one listing carries
histogram_quantile(0.95, sum(rate(mcp_hangar_projected_surface_bytes_bucket{kind="governed"}[5m])) by (le))

# What the surface is made of: bytes each upstream adds to a listing that includes it
sum(rate(mcp_hangar_projected_upstream_bytes_sum[5m])) by (mcp_server) / sum(rate(mcp_hangar_projected_upstream_bytes_count[5m])) by (mcp_server)
```

A change is counted when a caller lists again and is served something different
from what it was served before. The caller is the same tenant and principal,
and the same session when there is one. It covers a tool added, removed, routed
elsewhere or redefined, and the first listing after the warm-up lands. The
count is per replica and per caller. A caller's first listing on a replica is
not a change, and neither is the first listing of a caller the replica stopped
remembering. The memory holds 1024 callers, least recently served forgotten
first. So the counter can miss a change, but it never counts one that did not
happen. A client behind a load balancer without affinity may see a change
that no replica counted.
````

## Risk and rollback

Low. This is additive observability. What the client receives is unchanged: the listing's result, the call path and the error shapes are all untouched.

Per real listing, the new work is one compact JSON dump of each served definition and a SHA-256 over them. `generate_projection` already runs `model_validate` on every definition on every listing, which costs more.

Memory is bounded at 1024 × 32-byte digests.

New series: 2 × 13 for the surface bytes, 13 per upstream, and 1 for the counter.

To roll back, revert the single commit.

## Agent metadata

- [x] LOC declared upfront: about 200 in `src/`, excluding tests; about 1000 in tests, of which about 250 is the extracted harness and about 330 the evaluator


mapyr:

- **What gets measured.** Only listings the client actually received are counted; the SDK's pre-dispatch `tools/list` on a `tools/call` moves nothing. All four observations come from one function outside the handler, which is at its complexity ceiling.
- **Churn.** It uses `Projection` equality from #1381, compared as a length-prefixed SHA-256 held in a 1024-identity LRU keyed like #1395's stale-name memory. It can miss a change after an eviction or a first listing, but it never counts a change that did not happen. The counter is seeded at zero so `increase()` sees the first change.
- **Cardinality.** No metric carries a tenant label (#895). The labels are `kind`, with two values, and `mcp_server`, which is the configured fleet.
- **Checked against real scrapes.** The documented PromQL runs against two `/metrics` scrapes of a running gateway through a strict evaluator (`tests/_promql.py`), where a selector that matches nothing is an error. There is also a live T0 test against `serve --http`.
- **Nit, not blocking.** The evaluator adds about 400 lines of test infrastructure. It has its own tests, and CI has no Prometheus, so it earns its place.

**Docs follow-up.** In the new Front Door Projection subsection, add one line saying that `/metrics` is unauthenticated by default and should not be exposed on a public front door. CI is fully green.

Issue #1398, opened 2026-09-13
fix(core): circuit_reset_timeout_s has no effect on a group


mapyr:
## Symptom

`circuit_reset_timeout_s`, which a group accepts (`McpServerGroup(..., circuit_reset_timeout_s=60.0)`), has no effect. An open group circuit never moves to HALF_OPEN after the timeout, however long it waits.

## Root cause

`CircuitBreaker` checks its reset timeout only inside `allow_request()`; that is the only place OPEN becomes HALF_OPEN once `reset_timeout_s` has passed (`domain/model/circuit_breaker.py`). `McpServerGroup` never calls `allow_request()`. It only uses `record_failure()`, `is_open`, `reset()` (from `rebalance()`) and, since #1383 and #1397, `record_success()`. The timeout is configured, stored and read by nothing.

Since #1383, an open group circuit closes once `min_healthy` members are back in rotation, so the missing half-open probe no longer wedges a group. But the option still reads as if it did something.

## What

Decide which it should be:

- **Honour it.** Consult the breaker, via `allow_request()` or an equivalent, where the group decides whether to take a request, so an open circuit half-opens after the timeout and a probe can close it. Check how that interacts with `_maybe_close_circuit()`.
- **Remove it.** Delete `circuit_reset_timeout_s` from the group and its config, with an upgrade note for anyone who set it.

## Files in scope

- `src/mcp_hangar/domain/model/mcp_server_group.py`
- `src/mcp_hangar/server/config.py` (group config)
- `tests/`

## Acceptance criteria

- [ ] The decision is recorded here.
- [ ] If honoured: a test shows an open group circuit half-opening after `circuit_reset_timeout_s` and closing on a successful probe.
- [ ] If removed: a config that sets the option fails loudly or warns, rather than being silently ignored.

## Severity

P3: a configuration option that does nothing. Found while fixing #1390.

mapyr:
**Decision (maintainer, 2026-09-14): remove it, and warn.**

A group circuit already closes through `_maybe_close_circuit()` when members come back into rotation. That happens through health checks (the worker runs every 60 s by default) or through successful calls. A half-open probe would be a second, competing way to close the circuit.

Remove `circuit_reset_timeout_s` from the group and its config. A config that still sets it logs a warning at load, rather than being ignored silently. Add an `UPGRADE.md` note.

PR #1408, opened 2026-09-14
feat(core): expose group circuit breaker state as a per-replica metric

### LINKED ISSUE #1357: feat(core): expose group circuit breaker state as a per-replica metric
## Background

Split out of #1315.

## Why

Two replicas can disagree about the same group at the same instant, and nothing makes the divergence visible:

```
mcp-hangar-...-949g4 -> circuit_open=True   is_available=False  healthy=0  in_rotation=[F,F,F]
mcp-hangar-...-kjpkq -> circuit_open=False  is_available=True   healthy=3  in_rotation=[T,T,T]
```

A client pinned to the tripped replica gets `NoAvailableMemberError` indefinitely while the group is healthy everywhere else. That is a partial outage no metric reports, because the other replicas look fine. Until the state is shared (tracked separately under #1315), the divergence must at least be observable.

## What changes

Expose the breaker state as a gauge, labelled by group, scraped per replica so the disagreement shows up as a spread across pods rather than a single blended number.

Worth pairing with a `max by (group) - min by (group)` example in the docs, since the alertable condition is "replicas disagree", not "circuit is open".

## Acceptance criteria

- [ ] A gauge reports circuit state per group, emitted on every replica, not only the leader.
- [ ] The gauge is registered in `_register_all_metrics()` and a test asserts it actually appears on `/metrics` — see #1059, where five metrics were defined, incremented, documented with ready-made PromQL, and never registered, so every query in the project's own guide returned empty on every deployment.
- [ ] A test drives a group's circuit open and asserts the exposed value changes.
- [ ] Cardinality is bounded by group count; no tenant or member label.

## Out of scope

Changing breaker behaviour, and sharing state across replicas.

## References

- https://github.com/mcp-hangar/mcp-hangar/issues/1059

mapyr:
## Why

Each replica keeps its own group circuit breaker, so two replicas can disagree about one group. A client pinned to the replica whose circuit is open gets `NoAvailableMemberError` while the group serves from every other replica, and no metric showed it. This change exposes the breaker per replica, so the divergence is visible in Prometheus until shared state lands (#1358). Per the decision on #1380, the fleet view is PromQL, not an endpoint.

Closes #1357

## What

- **New gauge `mcp_hangar_group_circuit_open{group}`.** It reads 1 while this replica has the group's circuit open and 0 otherwise, and it is registered in `_register_all_metrics()`. Its only label is `group`: there is no tenant or member label. The scrape's `instance` label is what makes it per replica.
- **0/1, not a closed/half-open/open enum.** A group never half-opens its circuit (#1398). The breaker half-opens only in `allow_request()`, which the group never calls. A half-open breaker is one the group reports as closed and routes through, so it would read 0, just as `circuit_open` in `hangar_group_list` does. So the gauge always agrees with `hangar_group_list` on the same replica, and both documented queries below reduce to a max and a spread.
- **Where it is written.** From the breaker's state-change callback, which every transition passes through: opening on failures in a row, closing at `min_healthy`, and the reset in `rebalance()`. `McpServerGroup.observe_circuit(listener)` gives the listener the state immediately, then again after every transition, under the group lock. The server layer attaches the listener, in `server/bootstrap/group_circuit_metric.py`.
- **Why not from the bus.** Nothing publishes a group's events from the call path (the executor's `report_success`/`report_failure`) or from the health checks (`GroupRebalanceSaga`). So `GroupCircuitOpened`/`GroupCircuitClosed` never reach a bus subscriber from either. Putting breaker transitions on the shared log is also the design #1358 defers.
- **Seeded** when config loads a group at bootstrap (`_load_group_config`), so a series exists before the first transition. The listener is also attached to a group created through the API, from `GroupCreated`, as a `LOCAL_VIEW` handler.
- **Removed** on `GroupDeleted` by `remove_group_series`, the group counterpart of `remove_mcp_server_series`, called from `MetricsEventHandler` as an effect. The listener writes only while its group is the one `GROUPS` holds under that id. So a late transition of a deleted or replaced group does not put the series back. This narrows the race but does not close it.
- **Consecutive-failure count: considered, not exposed.** The count changes on every call through the group, so exposing it would put a gauge write on the call path instead of on transitions. It also resets on every success, so a 15–30 s scrape mostly samples 0 and can't reliably show the approach to the threshold. The transition is the signal worth alerting on.
- **The startup restore of group circuit state is untouched.** #1388 removes it, and nothing here builds on it.

## How tested

- `tests/unit/test_a_group_circuit_is_on_metrics.py` covers:
  - the listener hears every transition: failures in a row, the `min_healthy` close, and `rebalance()`;
  - the gauge is registered, typed `gauge` and labelled `group` only;
  - it is seeded, follows the circuit both ways and leaves on `GroupDeleted`;
  - a deleted or replaced group does not write it back;
  - a group from `GroupCreated` is observed.
- `tests/integration/test_each_replica_exposes_its_group_circuit.py` runs on `_group_recovery_harness.py`. That harness uses the real `bootstrap()`, `hangar_call` through the served app and the health worker, and it scrapes the `/metrics` endpoint that `serve --http` mounts. The tests check that:
  - `mcp_hangar_group_circuit_open{group="math-pool"} 0.0` is on the scrape right after bootstrap, before any call;
  - after every failing call the gauge equals `circuit_open` from `hangar_group_list`. In `single` it goes 0 → 1. In `pair` it goes 0 → 1 → 0 → 1, because math-b's start closes the circuit and its own failures open it again;
  - it reads 0 again once a passing health check returns a member to rotation;
  - it is gone after `DeleteGroupCommand` and back at 0 after `CreateGroupCommand`;
  - the documented PromQL below gives the expected answers when run through `tests/_promql.py` (#1406) against two real scrapes labelled as two replicas. The cases are one tripped and one recovered, both tripped, and both recovered.
- Local gates, Python 3.11:
  - `ruff check`, `ruff format --check`, `mypy` (436 files), `lint-imports` (1 contract kept), the dead-symbol baseline and C901 are all clean.
  - Unit (`pytest --ignore=tests/integration`): 8378 passed, 66 skipped, 1 failed. The failure is `test_there_are_dockerfiles_to_check`, which always fails under `.claude/worktrees` and is environmental.
  - Integration (`pytest tests/integration/ --timeout=60`): 458 passed, 5 skipped.
  - decision-coverage: CI's selection (`tests/unit tests/integration -m "not benchmark and not live"`) with coverage, run as a unit pass and an integration pass with `--cov-append`. Then `check_decision_coverage.py`: all 31 decision-path modules hold their floor, with `server/tools/batch/executor.py` at 93.33 against a floor of 90.5.

## Docs follow-up

For mcp-hangar/docs, add these to `guides/OBSERVABILITY.md` (the metrics table and the queries) and to the circuit breaker section of `guides/MCP_SERVER_GROUPS.md`. The queries live in `tests/_promql.py` as `GROUP_CIRCUIT_QUERIES`, and the guides must copy them verbatim.

Metrics table row:

| Metric | Type | Labels | Meaning |
| --- | --- | --- | --- |
| `mcp_hangar_group_circuit_open` | gauge | `group` | 1 while this replica has the group's circuit breaker open, 0 otherwise. |

Suggested text:

> Each replica keeps its own circuit breaker for a group, so replicas can disagree. A replica whose circuit is open refuses calls to the group with `NoAvailableMemberError`, while the others serve them. `mcp_hangar_group_circuit_open` is scraped from every replica, and the scrape's `instance` label tells them apart. Alert on disagreement, not only on an open circuit.

```promql
# Groups the replicas disagree about: open on at least one, closed on another.
max by (group) (mcp_hangar_group_circuit_open) - min by (group) (mcp_hangar_group_circuit_open) > 0

# Groups whose circuit is open on at least one replica.
max by (group) (mcp_hangar_group_circuit_open) == 1

# Which replicas have a group's circuit open.
mcp_hangar_group_circuit_open == 1
```

Notes for the guide:

- If several Hangar deployments share one Prometheus, add the label that separates them (for example `job` or `namespace`) to each `by (...)`.
- A replica that has not loaded the group has no series and does not count.
- A replica that is down drops out once its series go stale.
- For an alert, give the disagreement query a `for:` clause (for example `5m`), so a transition that one scrape catches mid-flight does not page.

## Risk and rollback

Low. The change is additive: one new gauge, and the group's routing and breaker behaviour are unchanged. The listener runs from the existing breaker callback, whose exceptions the breaker already swallows. To roll back, revert the single commit.

## Agent metadata

- [x] LOC declared upfront: 148 added, 3 removed in `src/` and `changelog.d/`; 549 added, 9 removed across 11 files including tests and the harness


mapyr:
The first `decision-coverage` run failed on `test_the_failed_restarts_count_against_the_member_until_it_leaves_rotation`, a #1399 test this PR doesn't touch. It passed 6 of 6 runs locally on `main` and 6 of 6 on this branch, and the re-run is green. The flake is tracked in #1411.

Issue #1411, opened 2026-09-14
test(core): make the dead-server failover test independent of runner speed


mapyr:
## Symptom

`tests/integration/test_a_given_up_server_reads_dead_on_the_served_path.py::test_the_failed_restarts_count_against_the_member_until_it_leaves_rotation` failed once on CI. It was PR #1408's `decision-coverage` job, run 34817798422:

```
AssertionError: [None, 'McpServerStartError', None, 'CircuitBreakerOpen', None, None, ...]
assert 3 == 2
 +  where 3 = len([1, 3, 14])
```

The two expected failures happened at calls 1 and 3. A third came at call 14. The error type of the third was cut off by pytest's repr truncation.

It does not reproduce locally. The test passed 6 of 6 runs on `main` (205cfdf2) and 6 of 6 on #1408's branch, whose change doesn't touch restarts, rotation or the executor. So it looks like a timing dependence that a slower runner exposes.

## Likely cause (not established)

The `failover` mode of `tests/integration/_dead_server_harness.py` runs every threshold at its default and then sends 20 calls through the group on wall-clock time. Something time-based can land inside those 20 calls on a slow runner. Candidates:

- member B's restart backoff elapsing, so that a call revives B and its failed start counts again;
- the saga's scheduled restart;
- a health check;
- a call to member A timing out.

## What

- Make the assertion message show every outcome's error type, so the next failure names its cause. Build the list explicitly rather than relying on the repr.
- Take the harness's timing out of the result: freeze or widen B's backoff for the window, or drive the calls against a controlled clock. Then the test proves the counting rule, not the runner's speed.

## Files in scope

- `tests/integration/_dead_server_harness.py`
- `tests/integration/test_a_given_up_server_reads_dead_on_the_served_path.py`

## Acceptance criteria

- [ ] The failure message lists all 20 outcomes with their error types.
- [ ] The test's outcome does not depend on how long the 20 calls take, and this is shown by running it with an artificial delay between calls.
- [ ] Single PR, test-only.

## Severity

P2: an intermittent red on a required job blocks unrelated PRs.

PR #1414, opened 2026-09-14
fix(core): remove the group circuit-breaker restore and the save it read

### LINKED ISSUE #1388: fix(core): group circuit state is never restored at startup, and would be restored to the wrong group

mapyr:
## Why

Closes #1388. The startup restore of group circuit-breaker state never restored anything. `init_saga()` ran it over `ctx.groups` while that was still an empty dict, and it loaded one row for every group rather than each group's own. The maintainer decided on the issue to remove it rather than fix it. Since #1383 a group's circuit closes by itself once its members recover, and breaker state stays local to each replica (#1380, #1358). Restoring a circuit saved as open hours earlier would hold traffic off a group that is healthy now.

## What

- Delete `_restore_group_circuit_breakers` and its call in `init_saga()` (`server/bootstrap/cqrs.py`).
- Delete `save_group_circuit_breakers` and the `ApplicationContext.shutdown()` hook that called it. Nothing writes `saga_type="circuit_breaker"` rows any more.
- Delete `CircuitBreaker.from_dict`, whose only caller was the restore, and `ApplicationContext.saga_state_store`, whose only reader was the save. `init_saga()` now returns `None`.
- `ServerLifecycle.shutdown()`: drop the comment, which appeared twice, that justified releasing the lease after `context.shutdown()` by the circuit-breaker row. The ordering stays, for the reason the comment after it already gives.
- `McpServerGroup._end_failure_run`: the docstring named a restored snapshot as the one way into HALF_OPEN. There is none now.
- Delete the tests that existed only for the removed code:
  - the restore and save tests in `test_saga_state_store.py` and `test_what_is_local_says_so.py`;
  - the `from_dict` tests in `test_circuit_breaker.py`;
  - the restored half-open test.
- The lease-ordering check moves from source inspection in `test_what_is_local_says_so.py` to a behavioural test in `test_lifecycle.py`.
- Add `tests/integration/test_a_leftover_circuit_breaker_row_is_ignored.py` and its harness.

**Rows left by older versions.** A deployment that ran an older version still has `circuit_breaker` rows in `saga_state`. Nothing loads that saga type now. `load()` is keyed by saga type on both SQLite and PostgreSQL, so the recovery and failover sagas' rows load as before. There is no migration and no `UPGRADE.md` note, because operator-visible behaviour does not change: every group already started closed. The changelog fragment says an old row is safe to leave or delete.

## How tested

- The new integration test runs the real `bootstrap()` in a fresh interpreter, on the SQLite persistence backend. Its `saga_state.db` already holds an OPEN `circuit_breaker` row under a configured group's id, next to a recovery saga checkpoint. After bootstrap it calls `ServerLifecycle.shutdown()`. It checks that:
  - both groups start closed with a failure count of 0: `math-pool`, which has the old row, and `spare-pool`, which has none;
  - the recovery saga's checkpoint still loads;
  - nothing raises;
  - after shutdown, the only `circuit_breaker` row is the old one, with the same state and `updated_at`, and no group gets a new row.
- Against `origin/main`'s `cqrs.py` and `bootstrap/__init__.py`, run from a scratch copy, the shutdown check fails: the old hook replaced `math-pool`'s open row with a closed one and added a row for `spare-pool`. The other four checks pass there too, because the restore never ran.
- Gates, run on the branch after rebasing onto main with #1408 merged:
  - `ruff check src/mcp_hangar`, `ruff format --check src/mcp_hangar`, `mypy src/mcp_hangar`, `lint-imports --config .importlinter`, `scripts/check_dead_symbols.py` (baseline holds) and `scripts/build_changelog.py check` all pass.
  - `pytest --ignore=tests/integration`: 8367 passed, 66 skipped, 1 failed. The failure is `test_there_are_dockerfiles_to_check`, which always fails under `.claude/worktrees`; it is environmental.
  - `pytest tests/integration/ --timeout=60`: 463 passed, 5 skipped.
  - Decision-coverage selection (`-m "not benchmark and not live"`, with coverage): 8706 passed, 9 skipped. `check_decision_coverage.py` reports that all 31 decision-path modules hold their floor.

## Risk and rollback

Low risk. The removed read never ran, and the removed write only wrote rows that nothing read. Revert the squash commit to restore both.

## Agent metadata

- [x] LOC declared upfront: +258 / -386 across 13 files, 6 of them test files
- [x] Files in scope match the issue brief. The brief names `cqrs.py` and tests. The decision to also delete the persistence the restore read reaches `bootstrap/__init__.py` (the shutdown hook), `circuit_breaker.py` (`from_dict`), and one comment and one docstring in `lifecycle.py` and `mcp_server_group.py`.


mapyr:
The `integration` failure is the known flake in `test_the_failed_restarts_count_against_the_member_until_it_leaves_rotation` (#1411). It's a test this PR doesn't touch, and it failed the same way on #1408. I've re-run the failed job.

PR #1417, opened 2026-09-14
fix(core): report group healthy_count as members healthy, not members in rotation

### LINKED ISSUE #1356: fix(core): report group healthy_count as members healthy, not members in rotation

mapyr:
## Why

A group's `healthy_count` counted every member in rotation that was not dead, `cold` ones included. A group whose members the GC had reaped reported them all healthy with nothing running, and with its circuit open it read `healthy_count: 3, is_available: false, circuit_open: true`. Both `GET /api/groups/{id}` and `hangar_details` reported it.

Closes #1356

## How tested

- New unit tests, `tests/unit/test_a_group_reports_ready_members_as_healthy.py` (7):
  - with every member `cold` and in rotation, `healthy_count` is 0 and `members_in_rotation_count` is 3;
  - the group stays available and selects a member;
  - the issue's shape with the circuit open is gone;
  - a cold member in rotation still counts toward `min_healthy`, so a success closes the circuit, and a crashed one does not;
  - `GroupStateChanged` carries both counts.
- New served-path test, `tests/integration/test_a_cold_group_reports_no_healthy_members_and_still_routes.py` (11), run by `_cold_group_harness.py`. It uses the real `bootstrap()`, the served MCP app and the REST API mounted at `/api`, the GC worker reaping idle members, and `hangar_call` through the group:

  | Moment | healthy | in rotation | total |
  | --- | --- | --- | --- |
  | booted | 2 | 2 | 2 |
  | both reaped to `cold` | 0 | 2 | 2 |
  | after one call | 1 | 2 | 2 |

  With both members reaped, the group reads `is_available: true`, `circuit_open: false`, state `healthy`. The call succeeds by starting one member. At every moment, `GET /api/groups/{id}`, `GET /api/groups`, `hangar_details`, `hangar_group_list` and `hangar_list` return identical dicts, and `hangar_status` and `hangar_health` report the same counts.
- Circuit behaviour is unchanged. Every #1397 test and every #1383 test except the one below pass untouched, as do the #1399 served-path tests and #1408's metric tests.
- Gates, after rebasing on #1412 and #1413:
  - `ruff check`, `ruff format --check`, `mypy`, `lint-imports`, dead symbols, and markdownlint on `UPGRADE.md` all pass.
  - Unit: 8308 passed, 1 failed before the #1412 fix below; that file passes after it (59 passed).
  - Integration: 469 passed.
  - `check_decision_coverage.py` on CI's selection: all 31 modules hold their floor.

### Existing tests that change

- **#1383, `test_idle_reaping_never_opens_the_circuit`:**
  - It asserted `healthy_count == len(IDLE_MEMBERS)` after the GC had reaped both members, so both were `cold`. That is the contradictory shape this PR removes.
  - It now asserts `members_in_rotation_count == 2` and `healthy_count == 0`. The circuit half is unchanged.
- **#1412, `GROUP_STATUS_KEYS` and `test_the_structured_group_entry_speaks_the_group_vocabulary`:** both pin the exact keys of a group entry, and they gain `members_in_rotation_count`.

## What

- **`healthy_count`:** counts the members that are `ready` and in rotation, and is only reported, never used for a decision.
  - Requiring rotation keeps healthy ≤ in rotation ≤ total.
  - It keeps #1383's "every member driven out" assertion true; those members are still `ready`.
- **`members_in_rotation_count`:** new. It counts the members in rotation in any state, the members `members_in_rotation` names. It is on `to_status_dict()` (REST, `hangar_details`, `hangar_group_list`, `hangar_list`), `hangar_start`, `hangar_group_rebalance`, `hangar_status`, `hangar_health` and `hangar_metrics`.
- **The group's decisions keep the old count, now under the private name `_live_rotation_count()`:** members in rotation that are not dead. `is_available`, the group state and the `min_healthy` rule that closes an open circuit use it, exactly as they used `healthy_count`. A group of cold members still routes.
- **`GroupView`:** built from one `to_status_dict()` read under the group's lock, instead of separately locked reads.
- **`GroupStateChanged`:** `healthy_count` takes the new meaning, and the event gains `members_in_rotation_count` (defaults to 0).
- **Metrics:** none reads a group's healthy count. `mcp_hangar_group_circuit_open` and `tests/_promql.py` only read the circuit.
- **Docs:**
  - a `UPGRADE.md` "Next" section, since `healthy_count` changes meaning;
  - a changelog fragment.

## Risk and rollback

A consumer that treats `healthy_count: 0` as "cannot serve" now sees it for idle groups that serve fine. `UPGRADE.md` points to `is_available` and `members_in_rotation_count`. Routing and circuit decisions are unchanged. To roll back, revert the single commit.

Docs follow-up in `mcp-hangar/docs`: `reference/tools.md` documents the group counts for `hangar_list`, `hangar_start`, `hangar_status`, `hangar_details`, `hangar_health`, `hangar_metrics`, `hangar_group_list` and `hangar_group_rebalance`.

## Agent metadata

- [x] LOC declared upfront: `+600 / -36` (13 files; about 390 of the added lines are tests)


mapyr:
Reviewed. The decisions (`is_available`, `_update_state`, the `min_healthy` close) keep the old count through `_live_rotation_count()`. `healthy_count` is only reported now, and the status tools read one `to_status_dict()` snapshot.

Non-blocking nit, older than this PR: the docstring examples for `hangar_group_rebalance`, `hangar_start` and `hangar_health` still show a group with state `ready`. That's the server vocabulary; a group reads `healthy`, `partial`, `inactive` or `degraded` (#1378). It can go in a later docstring pass.

PR #1419, opened 2026-09-14
fix(core): reschedule a recovery restart the server refuses for backoff

### LINKED ISSUE #1401: fix(core): the recovery saga's restarts are refused by the server's own backoff, and nothing retries
## Symptom

With default settings, the recovery saga's first restart of a degraded server is refused, and the saga never schedules another. The defaults are `max_consecutive_failures: 3` and a saga `initial_backoff_s` of 5.0. The server stays `degraded` until a call happens to start it. The saga's "retry with backoff, then give up" never runs.

Reproduced on `origin/main` (`f688e427`). The restart is run through `StartMcpServerHandler` 5s after the third failure, the same way `SagaManager` fires it:

```
saga first restart delay_s = 5.0
server backoff after 3 failures: min=7.20s max=8.80s
restart: CannotStartMcpServerError: Cannot start mcp_server: backoff not elapsed, retry in 2.6s
state after the refused restart: degraded | events recorded: []
```

The later retries have the same shape. The saga fires at 10s and then 20s, while the server's own backoff after 4 and 5 failures is about 16s and 32s.

## Root cause

- **The saga schedules restarts on its own clock.** `initial_backoff_s: float = 5.0` (`application/sagas/mcp_server_recovery_saga.py:37`). Retry *k* fires `5 * 2**(k-1)` seconds after a `McpServerDegraded` (`:180`).
- **The server refuses any start from `DEGRADED` inside its own backoff.**
  - `_can_start()` checks `if not self._health.can_retry():` (`domain/model/mcp_server.py:563`), and `ensure_ready()` raises `CannotStartMcpServerError("backoff not elapsed ...")` (`:610-614`).
  - The backoff is `base = min(60.0, 2**self._consecutive_failures)` with ±10% jitter (`domain/model/health_tracker.py:206-210`), which is 7.2–8.8s after 3 failures.
- **A refused restart records nothing.**
  - `StartMcpServerHandler` re-raises the error (`application/commands/handlers.py:91-95`).
  - `SagaManager._fire` catches it and logs `scheduled_command_failed` (`infrastructure/saga_manager.py:126-132`).
  - No event is recorded, and the saga acts only on events (`handled_events`, `mcp_server_recovery_saga.py:60-61`). Nothing schedules the next restart.

## What

The fix needs a choice, hence `status/triage`:

- The saga could never schedule a restart sooner than the server would accept one, for example by sizing its delay from the same backoff.
- Or a restart refused for backoff could be rescheduled rather than dropped.

Either way, a refused restart must not silently end recovery.

## Files in scope

```
src/mcp_hangar/application/sagas/mcp_server_recovery_saga.py
src/mcp_hangar/domain/model/health_tracker.py   (if the fix shares the backoff)
tests/unit/test_sagas.py
tests/integration/   (a served-path test with default settings)
```

## Acceptance criteria

- [ ] With default saga and health settings, a degraded server whose upstream stays down gets every restart attempt the saga schedules, and each attempt actually runs.
- [ ] A restart refused because the backoff has not elapsed does not end recovery.
- [ ] A served-path test (`bootstrap()`, the real workers and the saga) shows it with default settings.

## Severity

P2 -- the automatic recovery the saga provides does not run with default settings. A degraded server waits for traffic, and never reaches the give-up that #1361 makes visible. Starting it by hand or by a call works around it. It causes fewer restarts, not more, so it is not a resource problem.

Found while implementing #1359/#1361 (PR #1399).

mapyr:
## Why

Closes #1401. With default settings the recovery saga's first restart fired 5s after a degrade, inside the server's own ~8s backoff. The server refused it, `SagaManager._fire` only logged that, no event was recorded, and nothing scheduled another restart. The saga never restarted anything and never reached its give-up. Per the maintainer's decision on the issue, a restart refused for backoff is now rescheduled at the retry time the server reports. The refusal does not count as an attempt, and the saga does not copy the server's backoff formula.

## What

- **Where the retry time comes from.** `CannotStartMcpServerError` already carries it as a structured attribute, `time_until_retry`, which the model fills in from `health.time_until_retry()`. The fix reads that attribute and parses nothing from the message. It does not touch `domain/model/mcp_server.py`, `commands/handlers.py` or `domain/exceptions.py`.
- **`SagaManager.schedule_command(..., on_failure=)`** is a new optional hook, also declared on the `ISagaManager` port. When a scheduled command raises, the hook gets the exception and the manager sends the commands it returns. The hook runs under the manager's lock. A `cancel_all_scheduled_commands()` bumps a cancel epoch, so a timer that was already firing when shutdown cancelled it does not arm anything afterwards (#1389/#1407).
- **The recovery saga** answers a `CannotStartMcpServerError` from its restart, meaning no start ran, by scheduling the restart again:
  - The delay is the reported `time_until_retry` plus 0.25s.
  - It is never shorter than 1s and never longer than the saga's `max_backoff_s`.
  - A NaN or negative value is read as 0.
  - Any other error means a start ran and failed. That start already recorded `McpServerDegraded`, which counts the attempt as before, so the hook ignores it.
- **Loop guard:** `max_refusals` (default 20) refusals in one attempt count as one failed attempt. A server that keeps refusing, for example with a retry time of 0, still spends the budget and is given up on. A pathological server gets at most `max_retries x (max_refusals + 1)` timers. The server's jitter is re-drawn on every read, so a second refusal reporting ~0 is normal. At a 60s backoff, the 1s floor leaves at most about 12 of those.
- **Restart tokens.** A restart is registered before it is armed. A start, a stop or a give-up empties the server's list and cancels every token in it, rescheduled ones included. A refusal that arrives after that finds its token gone and is dropped. Each server has one live restart at a time: a new degrade supersedes a restart that is waiting after a refusal, so two restarts never run side by side. The saga's state now sits behind its own lock, which is never held while calling the manager.
- **Give-up** is unchanged: it follows `max_retries` real attempts and leaves the server `dead`/`given_up` (#1399).

## How tested

- **New unit tests:** `tests/unit/test_the_saga_reschedules_a_refused_restart.py`, 22 tests:
  - a refusal is rescheduled at the reported time plus the margin, and is not counted;
  - it is not rescheduled after a start, a stop or a give-up;
  - a pending reschedule is superseded by the next attempt;
  - a start that ran and failed is counted once;
  - the give-up follows `max_retries` real attempts even when each attempt was refused first;
  - the loop guard: 0, negative, NaN and inf retry times, a refusal count past the bound counting as an attempt, and an always-refusing server being given up on after a bounded number of timers;
  - on the real `SagaManager`: the hook receives the error and its commands are sent, a hook's reschedule is cancelled by `cancel_all_scheduled_commands`, a cancel during firing suppresses the follow-up, and a failing hook does not kill the timer thread;
  - an end-to-end refusal and retry on real timers.
- **New served-path test:** `tests/integration/test_the_saga_restarts_a_degraded_server_with_default_settings.py`, plus a `defaults` mode in `_dead_server_harness.py`.
  - It runs the real `bootstrap()`, the served app, the real health and GC workers and the registered sagas, with a real upstream that stays broken.
  - Every health setting is at its default. So is every saga setting except `max_retries`, which is 2 instead of 3. A third restart waits out a 32s server backoff, and the run would outlast the job's 60s pytest-timeout (8 + 16 + 32s plus detection is about 65s). With `max_retries: 2` the run takes about 33s.
  - A command-bus middleware that only watches records what each `StartMcpServerCommand` came to.
  - It shows that the first restart comes 5s after the degrade and is refused, that each attempt is refused first and then runs a real start, that there are exactly `max_retries` real starts followed by `dead`/`given_up`, and that nothing starts the server afterwards.
  - Against main's saga code the served-path test fails, because the saga never gives up. The new unit tests fail there too, since the hook they test does not exist on main.
- **Gates, run locally:**

  ```
  ruff check src/mcp_hangar                        # clean
  ruff format --check src/mcp_hangar               # clean
  mypy src/mcp_hangar                              # no issues
  lint-imports --config .importlinter              # 1 kept, 0 broken
  python scripts/check_dead_symbols.py             # baseline holds
  ruff check src --select C901                     # clean
  pytest --ignore=tests/integration                # pass (see note)
  pytest tests/integration/ --timeout=60           # 463 passed, 5 skipped
  pytest tests/unit tests/integration -m "not benchmark and not live" --cov=mcp_hangar --cov-report=json:coverage.json
                                                   # 8778 passed, 9 skipped
  python scripts/check_decision_coverage.py coverage.json   # all 31 modules hold their floor
  ```

  - `tests/ci/test_base_images_are_pinned_and_updated.py::test_there_are_dockerfiles_to_check` fails only when run from a `.claude/worktrees` checkout. That failure is environmental and was deselected locally.
  - The two known flaky tests, `test_the_failed_restarts_count_against_the_member_until_it_leaves_rotation` (#1411) and `test_a_call_after_its_backoff_revives_it`, passed on every run.

## Risk and rollback

Moderate risk. The saga now actually restarts degraded servers with default settings, so the number of restarts it issues goes up to what its configuration always claimed: at most `max_retries` real starts per degrade episode. Timing changes too:
- a degraded server is restarted, then given up on, instead of sitting `degraded` until a call;
- the saga keeps at most one restart timer per server alive at a time.

`schedule_command`'s new parameter is keyword-only and optional, so the failover saga and other callers are unchanged. To roll back, revert the single squash commit.

## Agent metadata

- [x] LOC declared upfront: 829 insertions / 51 deletions. About 260 lines are source changes in the saga, `SagaManager` and the port. The remaining ~570 are tests, harness and changelog.
- [x] Files in scope match the issue brief. The fix sits in the saga and `SagaManager`, as the task required. It also widens the `ISagaManager` port and adds a harness mode; `health_tracker.py` is not touched.


mapyr:
The `integration` failure is the known flake in `test_the_failed_restarts_count_against_the_member_until_it_leaves_rotation` (#1411). It's the harness's `failover` mode, which this PR leaves unchanged; the PR adds a separate `defaults` mode. The failed job is re-running.

Issue #1415, opened 2026-09-14
fix(core): apply a configuration passed to bootstrap() as a dict the same way as one read from a file


mapyr:
## Symptom

`bootstrap(config_dict=...)` does not apply every setting that loading the same configuration from a file applies. Settings that are applied only on the file path are silently ignored when the dict is passed directly:

- `tool_access.mode` (the topology). A dict that sets `front_door` comes up in egress mode.
- `interceptors`, which includes parameter validators.
- `ui_resources`, the `ui://` allow list.
- `param_validation`.
- `resource_links`.

The dict path also skips the config schema check, so an unknown or removed key gets no warning and no strict-mode refusal (seen while working on #1398).

The `Hangar` facade cannot express these settings, so this affects code that calls `bootstrap()` with a dict directly: embedders and test harnesses. For them, the configuration they passed and the configuration that runs differ, with nothing logged.

## Root cause

Only the file loader applies these sections: `load_config_from_file` and the helpers it calls. `bootstrap(config_dict=...)` goes through a narrower path. Find the exact split and list every section that differs.

## What

Make the two paths the same: one function applies a configuration dict, and both the file path and `bootstrap(config_dict=...)` call it. Whatever the file path validates, the dict path validates too, including the schema check and strict mode. If a section genuinely cannot be applied from a dict, refuse it loudly rather than dropping it.

## Acceptance criteria

- [ ] A parity test builds one configuration and boots it both ways: from a file, and via `bootstrap(config_dict=...)`. It asserts the same effective settings for every top-level section, including topology, interceptors, the UI allow list, param validation, resource links, and schema warnings or strict refusal.
- [ ] The test enumerates the sections from the schema, so a new section that only one path applies fails it.
- [ ] Changelog fragment. Say that embedders passing a dict now get the settings they passed.

## Severity

P2: configuration passed through a supported entry point is silently dropped.

Found while building served-path test harnesses.

Issue #886, opened 2026-08-11
front_door: two replicas serve different tools/list to the same tenant


mapyr:
Split from the `front_door` cold-projection parent.

Discovery is per-instance, and the flat projection is built from it, so two
replicas of one gateway answer the same tenant differently. Warm one of two,
then ask the same question with the same token:

```
  replica A  18 tools
  replica B   0 tools
```

and through the Service, alternating:

```
  18 tools
   0 tools
  18 tools
```

This is worse than a slow start. A client that lists tools, caches the answer
and calls one later can be told the tool does not exist — not because policy
changed, but because a different pod answered. For a product whose front door is
the tenant-facing surface, "what am I allowed to call" must not depend on
routing.

Note this compounds with the Service having no session affinity (filed
separately): those two together mean consecutive requests in one logical session
can hit different replicas *and* those replicas can disagree about the catalogue.

**Fix** is bound up with the sibling issue — if the projection's inputs stop
being per-replica lazy state, the divergence goes with it. Worth an explicit
test either way: two instances against one database, same tenant, byte-identical
`tools/list`.

mapyr:
Compounds with mcp-hangar/mcp-hangar#877: consecutive requests of one logical session can already land on different replicas, and this makes those replicas disagree about the catalogue as well.

mapyr:
## Independent verification — 2026-08-11

Confirmed against `main` @ `c4315a1`, same root cause as #885 and verified in detail there: the projection registry is fed **only** by discovery, per process (`application/event_handlers/tool_projection_handler.py:42-59` on `McpServerStarted`), and `front_door`'s `tools/list` is a per-request read of it (`fastmcp_server/flat_tool_projection.py:280` → `_build_flat_map`, l. 150). Two replicas that have started different sets of servers necessarily answer differently. Divergence is guaranteed by construction, not a race.

The proposed test — *two instances against one database, same tenant, byte-identical `tools/list`* — would fail today for a second reason worth knowing before writing it: the response carries a per-tenant `_meta.cacheScope` (`flat_tool_projection.py:132-147`), which is stable per tenant, so it will not break byte-identity; but when the tenant is unknown it is a fresh `uuid4` per request. Assert against a *known* tenant, or the test fails for the wrong reason.

**One correction, detailed in #878:** ADR-020 does not put `tools/list` on the projection side — it explicitly excludes the tool catalogue. What it *does* say, and what this issue actually disproves, is:

[...]

Issue #885, opened 2026-08-11
front_door: a cold replica cannot be warmed by any ordinary tenant


mapyr:
Split from the `front_door` cold-projection parent.

After a restart, a `front_door` replica has discovered nothing, so the flat
projection is empty, so `tools/list` returns `[]` to a valid tenant — and there
is no client-reachable action that changes this:

- `hangar_warm` and the rest of `hangar_*` are deliberately not projected in
  `front_door`, so the meta-API is unavailable by design;
- `tools/call` with a name the client already knows is refused by the same empty
  projection (`Tool 'add' not found`), so knowing the answer does not help;
- health checks do not warm a server that was never started —
  `health_check_interval_s: 30` was configured and both servers were still
  `cold/unknown` two minutes later.

Only `POST /api/mcp_servers/<id>/start`, per replica, fixes it.

**Possible fixes**, roughly in order of how much they change:

1. Start configured mcp_servers eagerly at boot in `front_door` mode — the
   projection is the served surface there, so its inputs cannot be lazy.
2. Have the flat `tools/list` warm cold servers on demand (bounded, and never
   blocking the response past a deadline), rather than projecting only what is
   already warm.
3. An `auto_start` per mcp_server spec that means what it says for standalone
   servers, not only for group members.

The readiness probe should probably not gate on this either way — `lifecycle.py`
already explains why gating readiness on a warm backend deadlocks (no endpoints
→ no calls → nothing warms). That reasoning is exactly why the *client-facing*
surface must not depend on warmth.

---

## Update 2026-08-12: the first bullet is no longer true, and it does not close this

`main` @ `708e7c2`. The premise that `hangar_*` is deliberately unprojected in `front_door` was overtaken by **#912** (`fbeeca2`, *project the management tools a front-door caller may call*, closes #904):

> The projection is now a property of the caller … a management tool is listed exactly when the caller may call it, from the same `TOOL_PERMISSIONS` table and the same authorizer.

`hangar_warm` requires `mcp_servers:lifecycle` (`server/tools/tool_permissions.py:65`) and is in that set. So a front-door caller holding lifecycle **can** now list and call `hangar_warm` over MCP. That is a real change to this issue's first bullet.

It is not a fix for this issue, for three reasons:

[...]

PR #1429, opened 2026-09-14
feat(core): bound tenant admission and reconcile replica readiness

### LINKED ISSUE #1428: feat(core): bounded tenant admission and recoverable replica readiness
Agent task: implement opt-in per-process tenant concurrency and rate budgets before tool execution, opt-in catalogue readiness with background recovery of failed starts, and bounded graceful HTTP shutdown for the live Kubernetes production-hardening request. Preserve existing behaviour when budgets and catalogue requirements are absent. Scope: server config/schema, batch admission, catalogue readiness/lifecycle, regression tests and one changelog fragment. Implementation budget: under 250 lines excluding tests. Validate tenant isolation, quota release and refill, retry after startup failure, readiness and existing audit/redaction tests.

mapyr:
## Why

An authenticated tenant can occupy all execution slots, and a frontdoor whose backend was unavailable during startup needs a manual warm-up. These changes provide opt-in tenant admission limits and automatic catalogue reconciliation for the live Kubernetes installation. Closes #1428.

## How tested

118 targeted unit tests passed, covering tenant admission, catalogue recovery, readiness, concurrency, lifecycle, argument/log redaction and audit identity. Ruff passed. The downstream image retains production 2.19.1 dependencies and is validated on staging before production.

## What

- Apply bounded, nonblocking per-tenant concurrency and token-bucket budgets before tool gates.
- Reject missing or unknown tenant budgets when configured; keep the default behaviour unchanged.
- Gate initial readiness with MCP_REQUIRED_CATALOGUE and retry server starts every 30 seconds without invoking tools.
- Limit graceful HTTP shutdown to 90 seconds.

## Risk and rollback

Budgets apply per process, so deployments must account for replica count. Required catalogue names must exist or initial readiness stays false. A later provider outage does not evict an already discovered replica. Revert the deployment image and remove the opt-in settings to roll back.

## Agent metadata

- [x] LOC declared upfront: under 250 implementation lines, excluding tests


mapyr:
Review: not mergeable as is. It needs to be split and reworked. Findings are listed most severe first.

**1. The catalogue reconciler fights the server lifecycle.**
- `reconcile_catalogue` sends `StartMcpServerCommand` for every name in `MCP_REQUIRED_CATALOGUE` every 30 s, forever, including after those servers have started.
- `StartMcpServerHandler` calls `ensure_ready()`, which is a deliberate start. For a `dead` server that means no backoff and no "not revived" check (`_can_start` applies both only when `by_call` is set). So the loop:
  - revives servers the recovery saga gave up on (#1399, #1401), with no backoff;
  - relaunches the process of a server stopped by a capability block, every 30 s;
  - restarts servers the GC stopped for being idle, which defeats `idle_ttl_s`;
  - does all of this in `egress` too, where lazy start is the documented behaviour.
- **It reverses two documented decisions on `main` with no issue or ADR saying so.** `build_readiness_report` says readiness "deliberately does NOT require a warm backend", and `warm_the_front_door_catalogue` says warming is front-door-only and "deliberately not retried" (#878, #885, #886). If those decisions should change, that needs deciding first.
- **The configuration is an environment variable only,** outside `config.yaml` and the schema, so it gets no validation. A typo keeps readiness at 503 forever. It is also invisible to `mcp-hangar config check` and to #1415's parity.

**2. Tenant admission holds the slot across every gate, including the approval hold.**
- `acquire()` wraps `_execute_call_inner`, and every gate runs inside it: tool access, approval, withdrawal, validators and cold start.
- With `max_concurrency: 1`, one call held for approval blocks the whole tenant for up to `approval_timeout_seconds`.
- A call the tool-access policy refuses still spends a rate token, so a principal that may not use a tool can drain the budget of the principals that may.
- Taking the slot after the policy gates and before the execution slot would avoid both.

**3. CI is red on this PR's own changes.** The SQLite failure is #1440, now fixed on `main`, and not this PR's.
- `mypy`: 4 errors in `tenant_admission.py` (lines 43, 44, 47 and 56; `Any | None` and `str | None` where `float`/`str` is expected).
- `test_every_code_the_executor_writes_is_sorted_on_purpose`: `TenantQuotaExceeded` is a new refusal code, but it is neither in `DENIAL_CODES` nor built through the pipeline's `refuse`. The front-door call log would mislabel the verdict, and the refusal gets no metric.

**4. The graceful shutdown bound has no effect.** `timeout_graceful_shutdown=90` is hardcoded, and the Helm chart sets neither `terminationGracePeriodSeconds` nor a `preStop` hook. Kubernetes' default 30 s grace period ends the pod first. It needs to be configurable, and the chart needs a matching change.

**5. Tests and hygiene.**
- **Tests:** all mocks. The executor test patches `get_identity_context`, and nothing exercises the served path. Per-request and per-tenant controls here are verified over real streamable HTTP.
- **Description:** "Ruff passed / 118 targeted tests" doesn't cover mypy or the full suite, which fail.
- **Missing pieces:** there is no `UPGRADE.md` note and no docs for `execution.tenant_limits` or `MCP_REQUIRED_CATALOGUE`, and the changelog fragment isn't named after the issue.
- **Branch:** it is now 7 commits behind `main` and conflicts in `server/config.py`, since #1430 reshaped `_init_concurrency_from_config` / `apply_process_config`.
- **Labels:** #1428 is labelled `agent-friendly`, but it changes the executor's gate path.

**What is worth keeping:** the token bucket in `tenant_admission.py` is sound: validation, a non-blocking refusal, and a release in `finally`.

mapyr:
Closing, per the review above. The work is split into #1445 (per-tenant budgets: the token bucket from this branch is reused, but taken after the policy gates), #1446 (catalogue readiness, which needs a decision first) and #1447 (a configurable graceful shutdown, plus room for it in the chart). The branch stays, so the bucket can be salvaged.

Issue #1426, opened 2026-09-14
fix(core): stop restarting unchanged servers on every configuration reload


mapyr:
## Symptom

Every configuration reload restarts every subprocess server, even when `config.yaml` is unchanged. The reload reports each one in `mcp_servers_updated`, stops it, and rebuilds it.

## Root cause

`_config_differs` compares the stored config, which carries the default `resources: {"memory": "512m", "cpu": "1.0"}`, against the freshly parsed one, which has `resources: {}` when the file sets none. So an unchanged server always compares as different.

## What

Compare normalised configs, with defaults applied on both sides, so only a real change counts as an update.

## Acceptance criteria

- [ ] Reloading an unchanged file updates, stops and restarts no server. Tested for subprocess, docker and remote servers.
- [ ] Changing one field of one server updates only that server.
- [ ] Changelog fragment.

## Severity

P2: needless restarts, and dropped in-flight calls, on every reload. Reloads are now also triggered by the default file watcher.

PR #1430, opened 2026-09-14
fix(core): apply the whole configuration on reload, and keep the topology mode

### LINKED ISSUE #1424: fix(core): apply the whole configuration on reload, and keep the topology mode
## Symptom

A configuration reload applies only `mcp_servers`. Everything else the startup path applies is either reset or left stale until the process restarts. Seen on `main` `3a4cfd1e`, with `config.yaml` unchanged, for every reload trigger:

- `POST /api/config/reload`;
- the `hangar_reload_config` tool;
- SIGHUP;
- the config file watcher, which is on by default. A byte-identical rewrite of the file is enough.

**`tool_access.mode` reverts to `egress`.** After the reload, a gateway configured `front_door` resolves tool access with egress rules, and front-door-only behaviour stops applying. One example: its handling of callers that carry no tenant identity. A key issued without a tenant goes from an empty `tools/list` to the full list, and its calls succeed. With `allow_anonymous: true`, or auth off, the same holds for unauthenticated callers.

The tool surface is built at boot, so the gateway still looks like a front door. Nothing is logged when the mode changes.

**Other sections go stale.** Edits to these are never applied by a reload:

- `interceptors` (validators);
- `ui_resources`;
- `headers.param_validation`;
- `resource_links`;
- `execution`.

The running values stay as they were at boot. Deleting a section and reloading changes nothing.

**Runtime policies are wiped.** `clear_all()` also removes tool-access policies set at runtime, such as those set over REST. And between the clear and the re-registration there is a window with no policies at all.

**The docstrings describe something else.** `server/api/route_permissions.py:46-48` and `server/tools/tool_permissions.py:35-37` say a reload "re-applies every governance input -- tool-access policies, digest pins, topology mode".

## Root cause

- `ReloadConfigurationHandler.handle` (`application/commands/reload_handler.py`, step 4) calls `get_tool_access_resolver().clear_all()`.
- `ToolAccessResolver.clear_all()` (`domain/services/tool_access_resolver.py:709-721`) resets `_topology_mode` to the default.
- The reload then calls only `apply_mcp_servers(...)`.
- The topology and the other process-wide sections are applied by the `_init_*_from_config` calls, and only `load_configuration` makes them (`server/config.py:1151-1156`), at startup.

## Workaround until this is fixed

On a `front_door` gateway, set `config_reload.enabled: false` and restart the gateway to apply configuration changes. Do not use `POST /api/config/reload`, `hangar_reload_config` or SIGHUP.

## What

1. **`clear_all()` must not reset the topology mode.** Tests that need a full reset can use a separate `reset()`.
2. **A reload applies the whole configuration** through the same function startup uses: `apply_configuration` from #1415 / PR #1421, or its `_init_*` part. That covers topology, interceptors, UI resources, headers, resource links and execution.
3. **Changing `tool_access.mode` by reload is refused** with a clear "restart required" error. The front-door tool surface is built at boot and cannot follow a live change.
4. **Swap policies instead of clearing first.** Build the new policy set, then swap it in, so there is no empty window. Say what happens to runtime-set policies across a reload, and keep them unless the reload replaces them.
5. **Correct the two docstrings.**

## Acceptance criteria

- [ ] A live test: a `front_door` gateway, a key without a tenant, then `POST /api/config/reload` with an unchanged file. Afterwards `tools/list` is still empty and the call still answers `-32601`, as before the reload. The same holds for SIGHUP and the file watcher.
- [ ] A reload that changes `tool_access.mode` is refused and changes nothing.
- [ ] Each of `interceptors`, `ui_resources`, `headers.param_validation`, `resource_links` and `execution`: an edit applied by reload takes effect, and deleting the block and reloading removes it. Each is shown on the served path.
- [ ] Changelog fragment and an `UPGRADE.md` note.

## Severity

P1: a reload silently changes how a front-door gateway resolves tool access, and the documented behaviour says it doesn't.

mapyr:
## Why

Closes #1424. A reload reset the topology mode to `egress` and applied only `mcp_servers`. `ReloadConfigurationHandler` called `ToolAccessResolver.clear_all()`, which also reset the mode, and then registered the servers' policies again one by one. `interceptors`, `ui_resources`, `headers.param_validation`, `resource_links` and `execution` were applied only at startup, policies set at runtime were removed, and between the clear and the re-registration the policy set was empty.

Built on `apply_configuration` from #1421, which has merged; this branch is based on `main`.

## What

- **Topology mode.**
  - `ToolAccessResolver.clear_all()` keeps the topology mode. A new `reset()` is the full reset, and `reset_tool_access_resolver()` uses it.
  - A reload refuses a file whose `tool_access.mode` differs from the running mode, with `ConfigurationRestartRequiredError`, a `ConfigurationError` subclass. The REST API answers 409, and nothing is changed.
  - Startup and reload parse the mode with the same function, `configured_topology_mode`.
- **Every section, through startup's function.**
  - A reload applies every process-wide section through `apply_process_config`, the function startup now uses. It is exposed on `IConfigLoader`.
  - A block deleted from the file goes back to its default.
  - `ui_resources` replaces the policies on the process guard, so the consent gate bootstrap attaches survives a reload.
- **Build first, then apply.** `IConfigLoader` offers `check_process_config`, `prepare_mcp_servers` and `commit_mcp_servers`. `load_config` is `build_config` followed by a commit.
  - Before it stops anything, a reload checks the process-wide sections, builds every server, group and governance block, and reads the REST endpoint's stored policies.
  - A bad block anywhere, or a policy store that cannot be read, refuses the reload, and the running configuration is left as it was.
  - The store refusal is `ConfigurationUnavailableError`, which the REST API answers with 503.
  - After that the reload stops and removes servers, applies the sections, and commits. None of those steps fails on the file.
- **Unchanged servers are kept, not replaced.**
  - The build reuses the running server object when the file declares the server exactly as it was built, so the groups hold it too.
  - The running object must also still read as the file builds it. The REST update endpoint rewrites a running server's `env`, `description`, `idle_ttl_s` and `health_check_interval_s` in place. A server edited that way is rebuilt from the file and the edited one is stopped, as a restart would apply the file.
  - The comparison reads those fields off the aggregate rather than from `to_config_dict()`, which redacts `env`.
  - A running server the new configuration replaces counts as updated and is stopped first.
  - A reload used to put a fresh copy of every server in the repository and stop only the ones it counted as changed. An unchanged running server was replaced without a stop: its process kept running outside idle timeout, GC and shutdown, and the next call started a second one.
  - This also fixes that orphaning for unchanged top-level servers, which already happened on `main`.
- **No empty window.** The commit replaces what is running; it never clears it first.
  - It swaps in the policies (`adopt_config_policies`), the withdrawals and pins (`adopt_config_overlays`) and the `header_exposure` blocks (`adopt_header_exposure_policies`).
  - Then it adds the servers and replaces the group registry: it sets the new groups, then drops the ones the file removed.
  - The group registry is never emptied. The front door finds a member's group only there, so while it was empty the group's deny list, access policies and withdrawals did not apply to its members.
- **Runtime policies.**
  - The REST endpoint's stored policies are replayed after the file with `_replay_tap_policies`, the function startup uses, so a reload and a restart agree on a scope both define.
  - The store is read when the reload is prepared, and read again just before the swap, both outside the resolver lock. If the second read fails, the first read's rows are used, so a REST write made during the reload is not undone.
  - The rows are then applied under the lock, together with the adopt, from rows already read. No policy check waits on the database.
  - Any other runtime policy, such as one `hangar_load` set, is kept unless the file now defines the same scope.
  - A server the reload removes takes its policies with it, as `hangar_unload` does.
- **Inline group members** count as declared by the file. They are no longer stopped as removed servers, and the policies set on them at runtime are no longer stripped.
- **Housekeeping.**
  - The reload handler is split below CC 15, and the C901 baseline cap goes from 14 to 13.
  - The reload docstrings in `server/api/route_permissions.py`, `server/tools/tool_permissions.py`, `configure_interceptors` and `flat_tool_projection` are corrected.
  - The `domain/services/tool_access_resolver.py` decision-coverage floor goes from 94.1 to 96.0.
  - `UPGRADE.md` note and changelog fragment.

Nothing under `auth/` is modified. The replay passes `_replay_tap_policies` a snapshot of the rows as its store.

Out of scope: #1426. `_config_differs` is unchanged, so a reload still restarts unchanged servers whose file omits `resources`.

### Known limits

- The policies, withdrawals and pins, and `header_exposure` blocks are three overlays, each swapped under its own lock, one after another. A call made between two swaps can see one overlay from the new configuration and another from the old one. None of them is ever empty.
- `execution` is applied by building a new `ConcurrencyManager`, so calls already running hold slots in the previous one.
- A server's own `max_concurrency` is set on the running manager while servers are built, rather than staged with the rest. So a reload refused later in the build has already changed it (#1432).
- A group's inline member whose id is also a top-level server is replaced without a stop. This also happens on `main` (#1437).

## How tested

- **Unit**
  - `tests/unit/test_a_reload_applies_the_whole_configuration.py`, through the real handler and `ServerConfigLoader`:
    - an unchanged front door stays a front door;
    - a mode change is refused in both directions and changes nothing;
    - a bad section is refused before any server stops;
    - each of the five sections is edited and then deleted;
    - the consent gate survives a reload;
    - withdrawals, pins and `header_exposure` are lifted when deleted;
    - runtime policies are kept, replaced or retired;
    - a runtime policy on an inline member survives an unchanged reload;
    - a group with an inline member comes back with its policies.
  - **Unchanged servers:** an unchanged top-level server and an unchanged inline member, both really started on `tests/mock_provider.py`, are the same objects after the reload, in the same state, and the group holds the running member. A server whose entry changed in a field the change check does not compare is stopped before it is replaced.
  - **REST edits:** after a REST edit of `env`, `description`, `idle_ttl_s` or `health_check_interval_s`, a reload of the unchanged file stops the server and rebuilds it from the file. A server really started on the mock upstream restarts with the file's `env`, which it reports back in its tool listing.
  - **Stored policies:**
    - A store read that fails refuses the reload, with nothing stopped and nothing changed.
    - A failed second read falls back to the rows read when the reload was prepared.
    - A REST write that lands between prepare and commit still wins after the reload.
    - A policy check made during a 0.5 s store read, at prepare and at commit, takes under 0.1 s.
  - **No gap:** the reload is paused after the new configuration is built and before it is in force. A caller on another thread still sees the previous deny policy, withdrawal, pin and `header_exposure` block, and the group's own deny list and withdrawal on its member. The new server is not in the repository yet.
  - **Failed build:** one edit deletes every process-wide section and breaks one server block. The reload is refused, and the sections, server aggregates, group and all governance are exactly as before. No server is stopped.
  - `tests/unit/test_config_policies_are_adopted_in_one_step.py` covers the resolver swap and provenance rules.
  - The review probes pass: `rev1430_repro.py` (`window`, `failed`, `inline`), `rev1430_repro2.py` (`unchanged`, `replayfail`, `lockhold`) and `rev1430_repro3.py` (`restupdate`, `envrotate`).
- **Integration (runs in CI)**
  - `tests/integration/test_a_reload_is_served_whole.py` runs the real `bootstrap(config_path=...)` and the app composed as `run_http` composes it, over streamable HTTP, one interpreter per scenario.
  - Front door, auth on, an API key without a tenant: `tools/list == []` and the call answers `-32601`. That holds at boot, after `POST /api/config/reload`, after a real SIGHUP, after the file watcher sees a save, and after a mode change refused with 409.
  - Front-door sections on the served surface: `ui_resources`, `resource_links`, `headers.param_validation`, and `interceptors` on a flat call.
  - Egress: `interceptors` through `hangar_call`, and `execution` through `POST /api/config/export`.
- **Live (T0)**
  - `tests/live/test_t0_reload_keeps_the_front_door.py` runs against this working tree's `mcp-hangar`. It covers REST, SIGHUP, the file watcher, and a mode change refused with 409.

```text
MCP_HANGAR_LIVE_VERIFY=1 uv run pytest tests/live/test_t0_reload_keeps_the_front_door.py -m "live and t0"
```

- **Gates (all run locally):** `ruff check`, `ruff format --check`, `mypy`, `lint-imports`, `scripts/check_dead_symbols.py`, the complexity baseline test, markdownlint, the unit and integration suites, the live test, and decision coverage with its floors.
- **Known flakes:** the dead-server harness tests (#1411) and `test_a_call_after_its_backoff_revives_it` are timing-dependent and may fail on a loaded runner.

## Risk and rollback

Medium. The reload path is rewritten, and startup's `load_config` now builds servers and overlays aside and commits them at the end, with the same result on a first load. A reload that changes `tool_access.mode` now fails where it used to succeed, and so does a reload while the policy store cannot be read; both are intended and in `UPGRADE.md`. Revert the squash commit to roll back.

## Agent metadata

- [x] LOC declared upfront: about 1,200 in `src/`; the rest is tests, `UPGRADE.md` and the fragment


mapyr:
The `integration` failure is `test_a_call_after_its_backoff_revives_it` in `test_a_given_up_server_reads_dead_on_the_served_path.py`, the same harness as the known flake in #1411. This PR's own reload tests passed. The failed job is re-running.

mapyr:
This stays a draft for now. An independent review found cases the PR doesn't cover yet, and they are being fixed on this branch:

- group rules while a reload is in progress, and after a reload that fails;
- a reload that fails while building servers leaves the process-wide sections already applied;
- runtime policies on inline group members;
- how runtime policies and file policies on the same scope combine after a reload.

I'll mark it ready after re-review.

PR #1434, opened 2026-09-14
fix(core): make the Hangar builder emit the remote endpoint and discovery keys the gateway reads

### LINKED ISSUE #1423: fix(core): make the Hangar builder emit the remote endpoint and discovery keys the gateway reads
## Symptom

Two `Hangar.from_builder()` features produce configuration the gateway does not read. Both have silently done nothing:

- **Remote servers.** `add_mcp_server(..., mode="remote", url="http://...")` writes the address under `url` (`src/mcp_hangar/facade.py:153`). The gateway reads a remote server's address from `endpoint` (`server/config.py:601`; the schema lists `endpoint` at `config_schema.py:73`), so the server has no address. The builder's own docstring example uses this form (`facade.py:86`).
- **Discovery.** `enable_discovery(...)` writes `discovery.docker`, `discovery.kubernetes` and `discovery.filesystem` (`facade.py:257-263`). The gateway's `discovery` section is `{enabled, sources, auto_register, refresh_interval_s, security}` (`config_schema.py:124`), so discovery is never enabled.

Until #1415 (PR #1421), a dict passed to `bootstrap()` skipped the schema check, so nothing warned. After it, both log `unknown_config_key`, and strict mode refuses them. That surfaces the problem, but doesn't fix the builder.

## What

- Make the builder emit the keys the gateway reads: `endpoint` for remote servers, and a real `discovery` section with `enabled: true` and `sources` in the gateway's shape.
- Keep `url=` as the builder's argument name if that is the public API, and map it to `endpoint`.

## Acceptance criteria

- [ ] A builder-made remote server boots with its address and answers a call over the served path, against a stub HTTP upstream.
- [ ] `enable_discovery(...)` produces a config the gateway accepts under strict mode, with discovery enabled for the requested sources.
- [ ] A test builds every builder option and asserts the result passes the schema check under strict mode. Then a builder key the gateway does not read fails the build.
- [ ] Changelog fragment.

## Severity

P2: two documented builder features have never taken effect.

Found while working on #1415.

mapyr:
## Why

Two `HangarConfig` features never took effect. `add_mcp_server(..., mode="remote", url=...)` wrote the address as `url`, but the gateway reads `endpoint`, so the server booted with no address. `enable_discovery(...)` wrote `discovery.docker`, `.kubernetes` and `.filesystem`, but the gateway reads `discovery: {enabled, sources}`, so discovery never turned on. Under the facade, discovery also never ran, because `Hangar.start()` bootstrapped the orchestrator and never started it.

Closes #1423

## How tested

- `tests/integration/test_a_builder_config_takes_effect.py`: each case runs `_builder_served_harness.py` in a fresh interpreter under `HANGAR_CONFIG_STRICT=1`. `SyncHangar.from_builder` boots the real `bootstrap()`.
  - **remote**: a builder-made remote server, pointed at the stub HTTP MCP upstream from `_front_door_harness`, boots with its `endpoint`. It answers a `hangar_call` through `mcp_app_for_serving`, the app `serve --http` serves, and also answers `SyncHangar.invoke`. The upstream sees both calls.
  - **discovery**: `enable_discovery(docker=True, kubernetes=True, filesystem=[dir])` boots under strict mode. The orchestrator holds, and the registry lists, all three sources in `additive` mode with the filesystem `path`, and the orchestrator is running. Neither Docker nor a cluster has to be reachable. Without the `kubernetes` extra, a recording stand-in replaces the kubernetes factory, and the test checks the config entry it received.
  - **filesystem**: after start, the orchestrator is running and the discovery thread is alive. After stop, neither is.
- `tests/unit/test_facade.py`:
  - A test builds every builder option (all four modes, every server option, all three discovery sources, `max_concurrency`) and passes the result through `prepare_config` under strict mode.
  - A signature guard fails if a new builder option or method is not exercised.
  - The pre-fix remote key and the pre-fix discovery shape each fail `build()`.
  - Refusal tests cover each case in the list below.
  - Facade discovery start and stop are tested on a dedicated loop, including the failed-start path.
- Negative control: with the `url` to `endpoint` mapping removed, the remote harness exits non-zero. `build()` names `mcp_servers.store has unknown key(s) ['url']`.
- Gates, run locally:
  - ruff check and ruff format --check pass.
  - mypy passes.
  - lint-imports: 1 contract kept.
  - The dead-symbol baseline holds.
  - The complexity gate is ruff C901, included in ruff check.
  - markdownlint: 0 errors.
  - `pytest --ignore=tests/integration`: 8480 passed. The one failure is `test_there_are_dockerfiles_to_check`, which always fails under `.claude/worktrees`.
  - `pytest tests/integration --timeout=60`: 492 passed.
  - decision-coverage run: all 31 modules hold their floor. `test_a_given_up_server_reads_dead_on_the_served_path::test_a_call_after_its_backoff_revives_it` failed once under coverage load and passed on re-run. It is in the same timing-dependent harness as #1411.

## What

- Map the builder's `url=` to the spec's `endpoint`. `url=` stays the public argument. The builder has no auth, headers or TLS options, so there was nothing else to map.
- Emit `discovery: {enabled: true, sources: [...]}`, one source per requested type:
  - docker: `{type: docker, mode: additive}`
  - kubernetes: `{type: kubernetes, mode: additive}`
  - filesystem: `{type: filesystem, mode: additive, path: <dir>}`
- `build()` runs `validate_config`. A key the gateway does not read fails the build, and the builder stays open so the caller can fix it.
- Refuse, with `ConfigurationError`, each builder option that was stored and never read:
  - a second filesystem directory: the gateway keys sources by type, so a second one replaced the first
  - `enable_discovery()` with no source
  - `mode="group"`: no members can be declared
  - an option the mode does not read: `url` outside remote mode, `image` outside docker and container mode, `env` or `command` on a remote server
  - `set_intervals()`: no config key sets the GC or health-check interval
- `to_dict()` no longer carries the facade's `max_concurrency`. `HangarConfigData` loses `gc_interval_s` and `health_check_interval_s`.
- `Hangar.start()` starts discovery when the configuration enables it, and `stop()` stops it. A failed start shuts the bootstrapped context down. The loop helper (`start_discovery_loop` / `stop_discovery_loop`) was extracted from `ServerLifecycle`, which now uses it as well.
- Add a changelog fragment and an `UPGRADE.md` "Next" note.

## Risk and rollback

Moderate, for builder users. Calls that used to be silently ignored now raise, and discovery configured under the facade, from a file or the builder, now actually runs and registers the servers it finds. `UPGRADE.md` lists each of these. Revert the single commit to roll back.

Found, not fixed here: the facade never starts the GC and health-check background workers. So under `Hangar`, `idle_ttl_s` is not reaped and periodic health checks do not run. Also, the docker source's connection retry sleeps with a blocking backoff on discovery's loop, so a stop with Docker unreachable waits it out, for `serve` as well.

## Agent metadata

- [x] LOC declared upfront: ~650 (about 270 in src, the rest tests and docs)


mapyr:
The `decision-coverage` failure is `test_a_call_after_its_backoff_revives_it`: one failure out of 8,847, in the same dead-server harness as the #1411 flake. This PR doesn't touch that harness, and the floors are unaffected. The job is re-running. This PR stays a draft until a batch of other merges lands, then it goes to ready.

mapyr:
Reviewed the source: nothing blocking. Moving discovery start and stop into `start_discovery_loop` / `stop_discovery_loop` is clean, and `run_coroutine_threadsafe` queues onto a loop that isn't running yet, so the old `ready.wait()` isn't needed. Three small points, none of them blocking:

1. **`container` mode doesn't require `image`.** `_OPTION_REQUIRED_BY_MODE` has no `CONTAINER` entry, and the config value object (`domain/value_objects/mcp_server.py`, around line 385) only validates subprocess, docker and remote. So a builder container server with no image passes `build()` and fails only when the launcher checks the image at start (`infrastructure/launchers/container.py:205`). Adding `McpServerMode.CONTAINER: "image"` to that table would move the error to build time.
2. **`command` on a docker server: is it read?** `_OPTIONS_READ_BY_MODE` treats `command` as read for docker, but the value object sets `command=None` in docker mode. If the docker launcher doesn't read the spec's `command` somewhere else, the builder should refuse it for docker the way it now refuses other options a mode doesn't read. Worth a quick check.
3. **A failed discovery start leaves `self._context` set.** `Hangar._start_discovery` shuts the context down and re-raises, but `self._context` stays set. A caller that still calls `stop()` would run `ApplicationContext.shutdown()` a second time, and that method has no guard for it. Setting `self._context = None` after that shutdown would close it off.

This stays a draft until a batch of other merges lands.

PR #1459, opened 2026-09-15
fix(core): refuse podman, mode-less and inline group-member servers in a declared cluster

### LINKED ISSUE #1456: fix(core): the declared-cluster local-mode check misses podman, a missing mode and inline group members
## Symptom

A config with a `coordination:` block is supposed to be refused at load if it carries any child-process server, because only the lease holder can run one and every follower answers it as having no tools. Three kinds of local-mode server still pass the check:

- `mode: podman`;
- a server with no `mode` at all, which defaults to `subprocess` everywhere else;
- an inline group member (a `members:` entry with its own `command`/`mode`, not defined under top-level `mcp_servers`).

Such a deployment loads, and followers serve those servers with no tools.

## Root cause

`src/mcp_hangar/server/bootstrap/persistence.py`:

```python
_CHILD_PROCESS_MODES: frozenset[str] = frozenset({"subprocess", "docker", "container"})
...
        if isinstance(spec, dict) and str(spec.get("mode", "")).strip().lower() in _CHILD_PROCESS_MODES
```

- `podman` is missing from the set, although `mcp_server.py` launches it as a container.
- `spec.get("mode", "")` turns a missing mode into `""`, which is not in the set. The loader (`config.py`, `build_config`) and `McpServerConfig` default it to `"subprocess"`.
- Only top-level `mcp_servers` entries are read. `_load_group_members` builds an inline member from its own spec when the top level doesn't define it, so the loop never sees those.

## What

- Add `podman` to the set, or better, derive the set from `McpServerMode` so a new local mode can't be forgotten.
- Default a missing mode to `subprocess` and normalise it the way `McpServerMode.normalize` does.
- Also walk `members:` of every `mode: group` entry. An inline member (one not defined at top level) whose mode is local is an offender, reported as `<group>/<member>`.
- Report every offender at once, as the function already does.

## Files in scope

- src/mcp_hangar/server/bootstrap/persistence.py
- tests/unit/ (the existing tests for `refuse_local_modes_in_a_declared_cluster`)
- changelog.d/

## Acceptance criteria

- [ ] `mode: podman`, a server with no `mode`, and an inline local-mode group member are each refused under `coordination:`.
- [ ] `mode: remote` servers, and groups whose members are all remote, still load.
- [ ] A member defined at top level is reported once, not twice.
- [ ] No behaviour change without a `coordination:` block.
- [ ] Single PR, <= 150 LOC.

## Severity

P2 -- a declared cluster can still load a config its followers can't serve; it fails quietly at request time instead of at load.

mapyr:
## Why

Closes #1456. A config with a `coordination:` block is refused at load if it declares a server that runs as a child process of one gateway. The check read only top-level `mcp_servers` entries whose `mode` was written as `subprocess`, `docker` or `container`. It missed a server with no `mode`, which the loader builds as `subprocess`; group members defined inline under `members:`; and `mode: podman`.

## What

- Take the refused modes from the launcher's `LOCAL_MODES`, the set it refuses to start on a replica that does not hold the lease. The load-time and launch-time refusals now agree, and `podman` is covered.
- Default a missing `mode` to `subprocess`, as `build_config` and `McpServerConfig.from_dict` do. Keep the existing strip/lower-case matching.
- Walk every `mode: group` entry's `members:`, the way `build_config` resolves them since #1458:
  - A member whose id names a top-level server is that server, wherever the group is in the file. It is checked once, as that top-level entry.
  - Any other member is inline, and its own mode decides. A local inline member is reported as `<group>/<member>`.
  - An inline member that two groups share is one server, so it is reported once.
- Skip a member entry that names no server and does not say how to run one, such as `{"id": "x"}` with no server `x`. A typo in a member id then reads as the loader's "names no server" refusal, not as a cluster error. The check reuses the loader's own `_defines_a_server` from `server/config.py`, with its `McpServerMode.normalize` handling, so the two can't drift. Importing it causes no cycle: fresh interpreters importing `persistence`, `config`, `bootstrap`, `server` or the CLI first all succeed.
- Keep `LocalModeInDeclaredClusterError` and the report of every offender at once.
- Invert `test_a_server_with_no_mode_is_left_to_the_schema`. Its premise was wrong: the loader builds a mode-less server with an `endpoint` as `subprocess`, not `remote`.
- Add an `UPGRADE.md` section, `## Next — a cluster refuses mode-less servers and local inline group members`, at the top. Main's `## Next` sections are unchanged.

About `podman`: `McpServerMode` has no `podman` member, so such a config never loaded. It passed this check and then failed at `load_config` with `'podman' is not a valid McpServerMode`. It is now refused earlier, with the cluster error. No podman config that loaded before is refused.

## How tested

```text
mypy src/mcp_hangar                                    -> Success: no issues found in 439 source files
ruff check / ruff format --check src/mcp_hangar        -> clean (CI's scope; see note)
pytest tests/unit -q                                   -> 9063 passed, 4 skipped
pytest tests/integration/{replica, lease, fencing, schema, tail, audit-bootstrap, front-door catalogue tests}
                                                       -> 48 passed, 5 skipped (postgres extra not installed locally)
scripts/build_changelog.py check <fragment>            -> OK
markdownlint-cli2 UPGRADE.md <fragment>                -> 0 issues
```

Coverage of the acceptance criteria in `tests/unit/test_a_declared_cluster_refuses_a_child_process_server.py`:

- `podman`, a mode-less server, and inline local and mode-less members are refused.
- Remote servers, and groups whose members are all remote (top-level or inline), still load.
- A group listed before a top-level remote server with the same id as its member is accepted.
- A member defined at top level is reported once, in any order, even when two groups name it. An inline member two groups share is reported once.
- Under `coordination:`, `{"id": "x"}` with no server `x` fails with the loader's `ConfigurationError` ("names no server") when run in bootstrap order (this check, then `load_config`), not with `LocalModeInDeclaredClusterError`.
- An inline member with `command:` is still reported as `g/x`.
- A guard fails if `McpServerMode` gains a mode that is neither `remote` nor `group` and is not refused here.
- Without `coordination:`, an inline local member still loads.

The only config in the repo with a `coordination:` block is the one embedded in `tests/acceptance/ha-gateway.yaml`. It declares `mcp_servers: {}`, so this change doesn't affect it.

Note: ruff over `src tests` flags one line in `tests/unit/test_tool_access_rules_is_named_as_a_key_nothing_reads.py` (E501, and a format change). That line came from main via #1451, and this PR does not touch it. CI - Core runs ruff on `src/mcp_hangar` only, which is clean, so main is green with it and I've left it alone.

`tests/ci/test_base_images_are_pinned_and_updated.py::test_there_are_dockerfiles_to_check` always fails when run under `.claude/worktrees`. That's environmental and unrelated to this change. `tests/ci` was not part of the local gates.

## Risk and rollback

A configuration with a `coordination:` block and a mode-less server or an inline local group member loaded before and is now refused at startup. The `UPGRADE.md` section covers the fix. Without `coordination:`, behaviour is unchanged. Revert the squash commit to roll back.

## Agent metadata

- [x] LOC declared upfront: about 110 lines of code and tests, plus the changelog fragment and the UPGRADE section
- [x] Files in scope match the issue brief (plus `UPGRADE.md`, for the newly refused configs)


mapyr:
Holding this as a draft until #1458 merges.

#1458 makes the loader build every top-level server before any group. A group member whose id names a top-level server is then always that server, whatever the order in the file.

This PR's group check follows today's order-dependent loader: with the group first, it counts the member as an inline entry. Once #1458 is in, that would refuse a group-first config whose member is a top-level remote server. After #1458 merges, this branch will take main and simplify the check: a member is the top-level server if one of that id exists anywhere, and inline otherwise.

PR #1477, opened 2026-09-15
fix(core): check task follow-ups against the current tool access, as calls are

### LINKED ISSUE #1473: fix(core): check task follow-ups against the current tool access, as calls are
## Symptom

A relayed upstream task is driven after the call that created it through `tasks/get`, `tasks/update` and `tasks/cancel` (`fastmcp_server/task_relay_handlers.py`). Those handlers check that the caller owns the task and that the session isn't suspended. They don't check the task's tool against the tool access policy in force now.

So if a tool is withdrawn, or the policy stops allowing it for the caller, after a task was created, the task's owner can still:

- poll it (`tasks/get`) and fetch its result (`tasks/result`);
- send it more input (`tasks/update`);
- cancel it (`tasks/cancel`).

A new call to the same tool is refused, so calls and follow-ups disagree about the same tool.

## What

- Record the task's server and tool with the task (the relay already keys tasks by owning server), and have each follow-up ask the same tool access resolver a new call asks: the tool policy, withdrawals and the tenant scope.
- Decide per method, and document it:
  - `tasks/update` sends new input, so it is refused when the tool is no longer allowed, like a new call.
  - `tasks/cancel` should stay allowed, since stopping work is always safe.
  - `tasks/get`/`tasks/result`: refuse or allow, per the design note in the PR. Suggested: refuse the result of a withdrawn tool, the same way a call's result isn't delivered.
- Use the same refusal code a call gets for the same reason.

## Files in scope

- src/mcp_hangar/fastmcp_server/task_relay_handlers.py
- the task registry, if it has to keep the tool name
- tests/unit/, and tests/integration/ for a served-path check
- changelog.d/, UPGRADE.md (a `## Next —` note)

## Acceptance criteria

- [ ] Over a real served path: create a task, withdraw its tool, then `tasks/update` is refused with the call's refusal code. `tasks/cancel` still works.
- [ ] The same when the policy stops allowing the tool for that caller or tenant.
- [ ] A follow-up for a tool still allowed behaves as today.

## Severity

P2 -- follow-ups and calls apply different rules to the same tool.

mapyr:
## Why

A relayed task's follow-ups (`tasks/get`, `tasks/update`, `tasks/cancel`) were checked for ownership and a suspended session, but not against the tool access in force now. Once a tool was withdrawn, or the policy stopped allowing it for the caller's tenant, a new call of the tool was refused while its task could still be sent input and polled for its result. Calls and follow-ups now get one answer for the same tool.

Closes #1473

## How tested

- **Served path** (`tests/integration/test_a_task_follow_up_is_checked_against_current_tool_access.py`, harness `_task_follow_up_access_harness.py`), on the front door and on egress: the real `bootstrap()` from a config file, API-key auth with two tenants, the served app under starlette's `TestClient`, and in-process upstreams whose every `tools/call` answers with a task. Each tenant creates a task with every tool. Then the file is reloaded through the reload command that `POST /api/config/reload` and `hangar_reload_config` send. The reload withdraws a group tool for every tenant and another for `tenant-a`, denies a group tool by policy, and denies a tool on an ungrouped server by `tenant-a`'s policy. Then:
  - `tasks/update` of each tool taken away is refused with `-32602`, the call's message and `data.error_type` (`ToolWithdrawnError` or `ToolAccessDeniedError`). The upstream is never sent the update.
  - On egress, `hangar_call` of the same tool is refused with the same `error_type`. On the front door it is `-32601`, because the tool is no longer listed.
  - `tasks/cancel` is served and relayed to the upstream for every task.
  - `tasks/get` still answers with the status.
  - Every tool still allowed, including `tenant-b`'s for the changes scoped to `tenant-a`, is updated and relayed as before.
- **Unit** (`tests/unit/test_a_task_follow_up_is_checked_against_current_tool_access.py`): each method's decision, including every output-carrying poll shape and that a legacy upstream's `tasks/result` is not fetched. Also covered: the check is asked about the call that created the task, a task without a recorded tool is refused, and `current_tool_access_refusal` itself (group and member scopes, owning groups, tenant withdrawal, the member projection fallback).
- `tests/integration/test_a_front_door_task_call_is_governed_by_its_group.py` still passes. Its harness's upstream now also answers `tasks/update` and `tasks/cancel`, which the new harness reuses.
- Gates on the final tree:
  - `mypy src/mcp_hangar`: clean.
  - `ruff check src/mcp_hangar` and `ruff format --check src/mcp_hangar`: clean.
  - `pytest tests/unit tests/integration -m "not benchmark and not live" --cov=mcp_hangar --cov-report=json:coverage.json` (this includes `tests/unit` and the task-relay integration tests): 9698 passed, 11 skipped.
  - `python scripts/check_decision_coverage.py coverage.json`: all 31 decision-path modules hold their floor (`server/tools/batch/executor.py` 95.91 against a floor of 94.0).
  - `test_there_are_dockerfiles_to_check` is known to fail in a `.claude/worktrees` checkout, where this was built. It did not fail in this run.

## What

- `GovernedTaskStore` records the server and tool of the call that created a task. `relay_and_govern` passes them through `register_relayed_task`, and `task_tool(key)` reads them back.
- `executor.current_tool_access_refusal(mcp_server, tool, tenant_id, target_server_id=...)` answers what a new call of the tool would be refused with now. It is built from the access and withdrawal gates' own pieces, now factored out: `_policy_scopes`, `_allowed_in_every_scope`, `_resolve_projection`, `_withdrawn_in_scope` and the refusal messages. The gates use the same helpers, so a follow-up and a call cannot ask different questions.
- `task_relay_handlers` asks it per method, as below.

### Per-method decisions

- **`tasks/update`** sends the tool new input, so it is refused like a new call, before the consent gate opens or the upstream is probed.
- **`tasks/cancel`** is never checked. Stopping work is always allowed.
- **`tasks/get`**, and the upstream `tasks/result` fetch behind it for an older upstream: a status-only answer is still served, so a client can watch a task it cancelled end. An answer that would hand over what the tool produced is refused, and the payload is never fetched for it. That means any `completed`, `failed` or `input_required` answer, or any `result`, `error` or `inputRequests` field.
- **Refusal shape**: JSON-RPC `-32602`, with the call's message and `data.error_type` set to the call's `error_type`. `tasks/*` have no tool-error envelope, so the call's code travels in `data`, the way the session-suspension refusal carries its reason. `-32602` is the code this relay already gives a follow-up it refuses on tool grounds (a drifted digest).
- **A task registered without its tool** is refused the same way (fail-closed). Every production registration goes through the relay seam, which records it.

## Risk and rollback

Low. A client following up a task whose tool it could no longer call now gets a refusal on `tasks/update` and on a poll that carries output. Cancels, status polls and tools still allowed are unchanged. Revert the single commit.

## Agent metadata

- [x] LOC declared upfront: `+929/-37` (about 240 in `src/`, 27 in docs, the rest tests)
- [x] Files in scope match the issue brief, plus `server/tools/batch/executor.py`, where the shared check lives


mapyr:
Holding this as a draft until #1476 merges. Both PRs extract the same policy-scope logic in `server/tools/batch/executor.py`, so they conflict in one block of `_CallPipeline.policy_scopes`.

After #1476 lands, the resolution is:
- keep this PR's `_policy_scopes(...)` call;
- have the non-group branch of `_policy_scopes` return #1476's `member_policy_scopes(mcp_server, owning_groups)`, so there is one helper for the member-named-directly case.

Issue #772, opened 2026-08-05
McpServerRegistered never reaches the event store: a server's birth is missing from its own stream


mapyr:
Found while verifying discovery end to end (2026-08-05, `main` at `0613fc7`).

## What happens

```
POST /api/mcp_servers/  ->  {"mcp_server_id":"crud-probe","created":true}
sqlite> select count(*) from events;        ->  0

PATCH /api/mcp_servers/crud-probe  ->  {"updated":true}
sqlite> select stream_id, event_type, stream_version from events;
mcp_server:crud-probe | McpServerUpdated | 0
```

The store is wired and working — the update lands. The **registration does not**,
so the stream for an aggregate starts at version 0 with `McpServerUpdated` and
the event that created it is nowhere.

## Why

`CreateMcpServerHandler` publishes through `EventBus.publish()`, whose own
docstring says:

> Note: This method does NOT persist events. Use publish_to_stream() for event
> persistence.

Sibling handlers in the same file use `publish_aggregate_events(...)`, which does
persist. So this is one call on one line, not a design decision — `McpServerRegistered`,
`McpServerDeregistered` and anything else published the same way are delivered to
in-process subscribers and then forgotten.

## Why it matters, and what it does not break

- **Audit**: "which servers exist, when were they added, and by which door" is
  the most audit-relevant fact this system holds, and it is the one fact the
  event store does not contain. `source` — `discovery:kubernetes`, a REST caller
  — is carried on the event and dropped with it.
- **Not replay**: `restore_from_events` deliberately takes identity and
  configuration from configuration rather than from the stream (#759), so a
  missing registration event does not corrupt a restore. This is an audit gap,
  not a correctness one.

## Adjacent

Worth auditing every remaining `event_bus.publish(` call site against the same
question, rather than fixing this one and leaving the pattern. A quick pass shows
several in `crud_handlers.py` alone (`EgressPolicyCleared`, `GroupDeleted`, ...).

Related: #762 (discovery emits no events at all), #753 (the ES spine).

PR #1526, opened 2026-09-20
fix(core): publish a group's events on the path that raised them, not on its next edit

### LINKED ISSUE #1410: fix(core): publish a group's events on the path that raised them, not on its next edit
## Symptom

Domain events a group raises outside the group CRUD commands are not published when they happen. That covers:

- a member leaving or rejoining rotation;
- the circuit opening or closing;
- a rebalance.

They collect on the aggregate, and go out together on the next group CRUD command. If none comes, they never go out, and a restart loses them.

## Root cause

The only places that drain a group's events are the CRUD handlers: `application/commands/crud_handlers.py:469`, `:522`, `:570`, `:626` and `:671`, each `publish_aggregate_events(MCP_SERVER_GROUP, group.id, group.collect_events())`. Nothing drains them after:

- `report_success()` or `report_failure()` on the call path (`server/tools/batch/executor.py:1165-1168`, `:1426`);
- the rebalance saga's calls (`application/sagas/group_rebalance_saga.py`);
- `rebalance()` through `hangar_group_rebalance`.

This is why #1357 writes its gauge from the breaker's callback rather than from `GroupCircuitOpened` and `GroupCircuitClosed`: subscribers to those events never hear them on time.

## What

Publish a group's events on the path that raised them, as the MCP server paths do. Drain after the executor reports a member outcome, after the saga acts, and after `rebalance()`. Then check what subscribes to group events (audit, metrics, alerts, the peer tail) for anything that assumed the old batching.

Decide, and state, whether these events belong on the shared log at all. #1358 defers sharing breaker state across replicas. Publishing is not sharing, but a peer projection must not start applying another replica's breaker transitions.

## Files in scope

- `src/mcp_hangar/server/tools/batch/executor.py`
- `src/mcp_hangar/application/sagas/group_rebalance_saga.py`
- `src/mcp_hangar/server/tools/groups.py` (rebalance)
- `tests/`

## Acceptance criteria

- [ ] A member failing out through calls publishes its events at that moment, and so does the circuit opening. A subscriber hears them before the call returns, shown on the served path.
- [ ] A health check that returns a member, and closes the circuit, publishes at that moment.
- [ ] No event is published twice, and nothing waits for a CRUD command.
- [ ] The decision about the shared log and peer projections is recorded here.

## Severity

P3: group events arrive late, in a batch, or not at all.

Found while implementing #1357.

mapyr:
## Why

Closes #1410. The only thing that drained a group's domain events was the group CRUD handlers, so a member leaving or rejoining rotation, the circuit opening or closing, and a rebalance reached subscribers late, batched with whatever else had piled up, or -- if nobody edited the group -- never, and a restart lost them. This is why #1357 had to write its gauge from the breaker's callback rather than from `GroupCircuitOpened`.

## What

- `application/group_events.py` is the one drain point: `publish_group_events(event_bus, group)`. Every path that makes a group record something calls it.
- Drains added where there were none: the executor after `_report_member` reports a call's outcome, `GroupRebalanceSaga` after it reports a health check to a group, `hangar_group_rebalance`, and the `POST /api/groups/{id}/rebalance` route. The saga takes the bus from `bootstrap()`; without one it behaves exactly as before.
- The five CRUD drain sites now go through the same helper instead of each calling `publish_aggregate_events` directly.
- `EventBus.publish_local` (also on `IEventBus` and the app-context protocol): delivers to this replica's handlers and appends nothing. See the decision below.
- Corrected two docstrings that asserted "nothing drains a group's events on the call path or in the health checks" -- true when #1357 was written, false now.

`_report_member` returns early for an `UNJUDGED` outcome: the group was told nothing, so it recorded nothing and there is nothing to drain.

## The shared-log decision, and peer projections

**Decision: publish locally, do not append these events to the shared event log.** The code does that, and a test pins it.

The split is by register, not by path:

- **A group's configuration** -- created, updated, member added or removed, deleted -- is one operator's decision, identical on every replica. Unchanged: it still goes to the group's stream on the shared log.
- **A group's rotation and circuit breaker** are this replica's own. Three replicas serving one group have three rotations and three breakers and they disagree routinely. "The circuit opened" is not true of the group; it is true of one pod. `REPLICA_LOCAL_GROUP_EVENTS` is that set, and it is published with `publish_local`.

Why not the log:

1. **#1358 is open, `human-required`, `status/blocked` and ADR-gated**, and it says of the option it defers that it "puts breaker transitions on the event log, which brings the commit-ordering and projection/effect taxonomy rules into scope". Writing them to the log *is* that mechanism. A P3 delivery bug is not where that gets decided, and log rows are not easily unwritten once they are in an operator's database.
2. **It makes the peer-projection hazard structural rather than conventional.** The tailer can only deliver what is in the log, so no peer can apply another replica's breaker transitions whatever anybody subscribes later. Today nothing would: no `PROJECTION` handler subscribes to any group event, and `deliver_tailed` is projection-only. That is a convention two lines of code could break; keeping the events out of the log cannot be broken that way.
3. **Nothing is lost.** Every subscriber these events have today -- logging, metrics, alerts, audit, the security handler -- is an `EFFECT`, and effects run only on the replica that produced the event. The log would carry them to no one. The audit and compliance exporters still receive them, now on time rather than never.

I checked what actually subscribes, for anything that assumed the old batching: the four `subscribe_to_all` effect handlers, plus `GroupCreated` -> `observe_created_group` (`LOCAL_VIEW`) and `GroupDeleted` -> `MetricsEventHandler`. None of them are ordering- or batch-sensitive, and none are projections. `MetricsEventHandler` has no entry for the rotation or circuit events.

Two consequences worth naming:

- A group records `CircuitBreakerStateChanged` carrying **the group's id in `mcp_server_id`**, and `MetricsEventHandler` maps that event to `update_circuit_breaker_state`. That series will now be written under a group id promptly instead of only when a CRUD command happened to follow. Pre-existing in kind, newly visible in practice. Left alone deliberately -- renaming or re-keying that event is not this issue.
- Events that config loading leaves on a group at bootstrap were previously drained by nothing; the first call through the group now flushes them. `GroupCreated`/`GroupMemberAdded` among them are configuration and correctly go to the stream. The served-path test reads its assertions as deltas for exactly this reason.

## How tested

Two new files, both of which fail without the fix (verified by neutering the four new drain sites and re-running; restored afterwards):

- `tests/unit/test_a_groups_events_go_out_on_the_path_that_raised_them.py` -- 17 tests. A real `McpServerGroup` driven through the real `BatchExecutor`; the saga; the rebalance tool; `publish_local` against a real `EventBus` + `InMemoryEventStore` asserting the group's stream stays empty while a subscriber hears everything, and that a CRUD command still writes the stream. Plus an AST walk over `McpServerGroup` asserting every event it records is sorted into exactly one of the two sets, so a new one has to be classified on purpose. **6 failed without the fix.**
- `tests/integration/test_a_groups_events_reach_a_subscriber_on_the_call.py` -- 7 tests on the served path, via a new `events` mode in `_group_recovery_harness.py`: real `bootstrap()`, `hangar_call` through the app `serve --http` serves, a group over stdio in front of `tests/mock_provider.py`, and the health-check worker. An ordinary `subscribe_to_all` subscriber; snapshots after every call show the member leaving rotation and the circuit opening arrived *on the call that caused them*, and the passing health check publishing the member back and the circuit closed at that moment. No CRUD command and no rebalance is sent. **4 failed without the fix.**

Gate results, all from a private venv inside this worktree (`mcp_hangar.__file__` verified inside it):

```
ruff check src/mcp_hangar tests            All checks passed!
ruff format --check src/mcp_hangar tests   1113 files already formatted
mypy src/mcp_hangar                        Success: no issues found in 443 source files
pytest tests/unit                          9109 passed, 274 skipped
pytest tests/integration                   714 passed, 49 skipped
python scripts/check_dead_symbols.py       baseline holds
lint-imports --config .importlinter        Contracts: 1 kept, 0 broken
python scripts/build_changelog.py check    OK: 1 changelog fragment(s) valid
pytest tests/ci                            102 passed, 1 skipped, 1 failed (see below)
```

`tests/ci/test_base_images_are_pinned_and_updated.py::test_there_are_dockerfiles_to_check` fails with `assert 0 >= 4`: the Dockerfile glob matches nothing under a `.claude/worktrees` checkout. Environmental, unrelated to this change, and expected to pass on CI.

## Risk and rollback

Low. Additive at the call sites; no aggregate behaviour changed, so rotation and circuit decisions are byte-identical. The one behavioural change is *when* and *where* already-existing events are published. Nine existing test doubles needed a `collect_events` that returns a list (or an `event_bus` on a `SimpleNamespace` ctx) now that these paths drain -- all test-double updates, no production assertions relaxed. Revert the single squash commit.

## Agent metadata

- [x] LOC declared upfront: ~190 excluding tests

Two files outside the issue's "Files in scope" list, both deliberate and named here: `server/api/groups.py` (the REST rebalance route raises the same events as the tool -- draining one and not the other would make whether an operator hears about it depend on which surface they used) and `infrastructure/event_bus.py` + the two bus interfaces (`publish_local` is what the decision above required).


mapyr:
## Review — one thing to settle before this merges

I checked the load-bearing decision and I agree with it: rotation and breaker events are facts about *this replica*, `publish_local` keeps them out of the shared log, and that is the conservative reading of #1358 while it is still open and ADR-gated. Keeping them unlogged makes the peer-projection hazard structural rather than conventional, which is the right trade. `tests/unit/test_publish_persists_by_construction.py` is untouched, and no `PROJECTION` subscribes to a group event today, so nothing loses delivery.

**The finding is about where `publish_local` lives, not what it does.**

It is added as an `@abstractmethod` on `IEventBus` (`domain/contracts/event_bus.py:115`) and on the `server/context.py` Protocol. That makes a non-persisting publish door available to every holder of a bus, repo-wide. Today it has exactly one production caller — `application/group_events.py:133` — and the module docstring says so, but nothing enforces it.

That is the exact shape #772 catalogued. Quoting the guard test this PR leaves in place:

> "Which method should I call" is not a question a caller can get right reliably, and getting it wrong is silent.

#772 was 34 call sites taking the forgetful door against 10 that did not. The new test pins **which events** may go local (the AST walk over `McpServerGroup`) — good, and it means a *new group event* must be classified on purpose. It does not pin **who may call it**. A future caller handing some other aggregate's event to `publish_local` gets silent non-persistence, with no test to catch it, and the docstring is the only thing standing in the way.

**Suggested fix, small:** a guard test asserting `publish_local` has exactly one call site under `src/` (AST or grep over the tree), in the spirit of `test_publish_persists_by_construction.py`. Roughly twenty lines, and it converts the docstring's claim into something CI keeps true.

A narrower alternative, if you prefer: drop `publish_local` from the abstract port and have `group_events` depend on the concrete `EventBus`, so the door is not on the interface every consumer sees. That is a bigger change and I would not insist on it.

Not blocking on my side beyond this — the rest of the change reads well, and the two-register split in `group_events.py` is the right seam.

PR #1531, opened 2026-09-20
chore(ci): enable ruff import sorting and fix the hook names

### LINKED ISSUE #1496: chore(ci): make the pre-commit ruff hook match CI, and decide on import sorting
## Goal

Make the pre-commit ruff hook and CI agree, and decide whether import sorting is enforced.

## Motivation

Found while working on #1481:

- The pre-commit hook is described as "lint + isort", but ruff's config doesn't select the `I` rules, so import order isn't checked anywhere.
- The hook pins ruff 0.14.13, while the dev environment resolves 0.16.7. The two can format differently.
- Running `ruff --select I --fix` rewrites 46 untouched files under `server/`, so turning it on is a one-time reformat.

## Proposed change

- Pin the pre-commit ruff to the version the dev extras resolve, and keep them in step (e.g. Dependabot for pre-commit, or a check in CI that compares them).
- Either:
  - add `I` to ruff's `select` and land the one-time reformat as its own commit, listed in `.git-blame-ignore-revs`;
  - or drop "isort" from the hook's description.

  Pick one.

## Acceptance criteria

- [ ] pre-commit and CI run the same ruff version.
- [ ] Import sorting is either enforced in both or described in neither.

mapyr:
## Why

Closes #1496. Import order was checked nowhere: `I` was absent from ruff's `select`, while the pre-commit hook was named `ruff check (lint, no import sorting)`, the CI step said the same, and the issue found the hook described elsewhere as "lint + isort". A name read as coverage that delivers none is worse than no hook. The issue said "pick one" -- this enables the rule rather than dropping the claim.

The version half of the issue was already satisfied on `main`: the hook `rev` and the dev-extra pin are both `v0.16.7`, guarded by `tests/unit/test_pre_commit_ruff_matches_the_dev_pin.py`. What was left was the decision.

## What

Three commits, deliberately separated so the reformat stays mechanical.

- **`c2fd3c35` (policy)** -- adds `"I"` to `select` in `[tool.ruff.lint]`; renames the pre-commit hook and the CI step to `ruff check (lint + isort)` and updates the comment blocks that asserted import order was unchecked; deletes `.github/.isort.cfg`; strengthens the guard test; adds the changelog fragment.
- **`1f87df85` (reformat)** -- **a pure mechanical reformat and nothing else.** It is listed in `.git-blame-ignore-revs`. See below.
- **`1f40ea91` (blame)** -- creates `.git-blame-ignore-revs` (the repo had none) and adds the one-line opt-in to `CONTRIBUTING.md`'s Quick Start and `AGENTS.md`'s Setup block.

### Commit `1f87df85` is a pure mechanical reformat

It contains no hand edits and no behavior change -- only the order of import statements. It was produced by exactly these two commands, using the ruff pinned in the dev extra (**0.16.7**, verified with `ruff --version` in a venv private to this worktree, since a reformat produced by a different ruff than CI runs would be rejected):

```
ruff check --select I --fix src/mcp_hangar tests
ruff format src/mcp_hangar tests
```

**583 I001 findings across 527 files** -- 230 findings in 217 files under `src/mcp_hangar`, 353 in 310 files under `tests`. `ruff format` afterwards reported *1113 files left unchanged*, so nothing in that commit is incidental reformatting.

(The issue's "46 files" figure was measured over `src/mcp_hangar/server/` alone, not the scope CI lints.)

### Nothing needed an exemption

No `# noqa`, no per-file ignore, no config carve-out. A 527-file import rewrite can break circular-import workarounds and imports placed after a side effect, so the order-sensitive sites were checked with `ruff check --select I --diff` *before* applying anything:

- Ruff sorts only within a **contiguous** block, so imports deliberately placed after code do not move. The nine files carrying `# noqa: E402` after a `sys.path` insert keep their position, as does the deferred `from .config import (...)` at the foot of `domain/value_objects/mcp_server.py` that breaks a cycle with `config.py`.
- `tests/unit/test_saga_state_store.py` asks for `application.commands` to be imported FIRST to break a circular chain. Sorting lifts it *above* the `infrastructure.persistence` imports that used to precede it, so the comment is now literally true.
- `tests/unit/test_tool_metadata_carried.py` needs `import mcp_hangar.server` ahead of the `from mcp_hangar...` imports; that is where ruff puts a straight import within a section anyway.

### The guard test was going vacuous

`test_nothing_claims_import_sorting_while_i_is_unselected` returned early once `I` was in `select` -- so this very change would have silently switched it off. Its docstring said "If `I` is ever enabled, the names may say isort again". It is now `test_the_names_say_what_ruff_actually_does_about_imports`: it reads the hook name and the CI step name and asserts the claim matches `select` **in both directions**, so the names cannot drift from the rule either way. The version-pin test is unchanged.

### `.github/.isort.cfg` deleted

Dead config from another project: `src_paths = registry`, `known_first_party = registry`, `profile = black`, `line_length = 88` against this repo's 120, and a comment citing a `lint.yml` that does not exist in `.github/workflows/`. Nothing read it, and leaving it while enabling ruff's isort would have left two contradictory answers to the same question.

## How tested

Every gate, against the pinned ruff in a venv private to this worktree (`mcp_hangar.__file__` verified to resolve inside it). Each was also run **before** the reformat, so the comparison is a real baseline rather than an assumption.

```
ruff check src/mcp_hangar tests          All checks passed!
ruff format --check src/mcp_hangar tests 1113 files already formatted
mypy src/mcp_hangar                      Success: no issues found in 443 source files
lint-imports --config .importlinter      Contracts: 1 kept, 0 broken.
pytest tests/unit                        9109 passed, 274 skipped
pytest tests/integration                 714 passed, 49 skipped
pytest tests/ci                          102 passed, 1 skipped, 1 failed (see below)
python scripts/build_changelog.py check  OK: 2 changelog fragment(s) valid
```

`tests/unit` and `tests/integration` are **identical before and after** the reformat (9109/274 and 714/49 both times). No regressions.

`tests/ci::test_there_are_dockerfiles_to_check` fails, and **fails identically on the unmodified baseline**. It is environmental -- the Dockerfile glob matches nothing under a git worktree -- and unrelated to this change.

The blame-ignore file was verified rather than assumed: line 12 of `domain/value_objects/mcp_server.py` is `import re`, which the reformat moved. `git blame` credits it to `1f87df85`; with `blame.ignoreRevsFile` set it falls back to `b41da342` from 2026-01-31, the commit that actually introduced it.

## Risk and rollback

Low. No runtime change -- import order is the only thing that moved, and the full suite is byte-for-byte equal in outcome. The lasting risk is the one this PR is paying for deliberately: `git blame` on 527 files now points at the reformat unless a contributor opts in, which is what `.git-blame-ignore-revs` and the two doc hints address.

Rollback: revert `1f87df85` (the reformat) and `c2fd3c35` (the `select` change) together. Reverting only one leaves the tree failing `ruff check`.

## Agent metadata

- [x] Single Conventional Commit scope: `ci`
- [x] LOC declared upfront: ~61 hand-written lines across policy + blame commits; the 527-file / 2395-line reformat is machine-generated and blame-ignored


mapyr:
## Independent check: commit 2 really is import-order and nothing else

The `.git-blame-ignore-revs` entry rests on `1f87df85` being purely mechanical, so I verified that claim by parsing rather than by reading the diff. For all **527** files in that commit I compared the two revisions with `ast`:

1. **the multiset of imports** — every `Import` / `ImportFrom`, as `(kind, module, level, name, asname)` — must be equal;
2. **the AST with every import statement removed** — `ast.dump` of the two must be identical, so any change outside the import lines shows up.

Result:

```
files proven: 527
import set changed     : 0 []
non-import code changed: 0 []
```

So no file gained, lost or renamed an import, and no file's non-import code moved. Nothing rode along with the reformat.

**One detail worth recording, because it nearly became a false alarm.** My first pass flagged `infrastructure/discovery/entrypoint_source.py` as having changed outside its imports. That was a bug in my checker, not in this PR: it stripped imports from `body`, `orelse` and `finalbody` but never descended into `ast.Try.handlers`, and that file's imports live in a `try:` / `except ImportError:` pair for the `importlib_metadata` backport. The actual change there is `from importlib.metadata import entry_points, EntryPoint` → `import EntryPoint, entry_points` — alphabetical, exactly what I001 does. Fixing the checker to strip imports from every statement-list field brings it to 0.

Two things I checked separately and can confirm from the tree:

- the deleted `.github/.isort.cfg` was not merely unused but *contradictory* — `src_paths = registry` and `known_first_party = registry` against a package called `mcp_hangar`, `line_length = 88` against the repo's 120, `force_sort_within_sections = true` against ruff's default, and a comment pointing at a `lint.yml` that does not exist in `.github/workflows/`. Deleting it removes a second, wrong answer to "how are imports sorted here";
- the guard test's new bidirectional form is the right shape. The old `test_nothing_claims_import_sorting_while_i_is_unselected` returned early once `I` was selected, so after this PR it would have asserted nothing at all.

Nothing blocking from me.

Issue #1539, opened 2026-09-21
IdentityContext carries no session id on streamable-HTTP, so nothing downstream can name the MCP session


mapyr:
## What happens

An API-key caller over streamable-HTTP has no session id in its
`IdentityContext`, so nothing downstream of the identity bridges can name the
MCP session a call belongs to.

Found while adding the live tier for #1278. The governance boundary exports
what the bound identity holds, and on a real request that is:

```
batch.call.add  mcp.caller.id=svc:tenant-trace  mcp.caller.tenant_id=tenant-trace
                mcp.caller.type=service         mcp.user.id=svc:tenant-trace
```

`mcp.session.id` is absent, and correctly so -- the contract is that an unknown
value is omitted rather than exported empty. But it is absent because
`CallerIdentity.session_id` is `None`, not because the request had no session:
the MCP session exists on the wire, and the SDK is holding one.

Reproduced on both surfaces (`egress` / `hangar_call` and `front_door` / flat
tool) in `tests/live/test_t3_governance_identity.py`, which asserts the absence
so the day it stops being true is a red test rather than a silent change.

## Why it matters

Three things key on a session and none of them can:

- a trace cannot group the calls of one client session, which is the question
  an operator asks first when a client misbehaves;
- `mcp.session.id` is in `conventions.py` and documented as the MCP session
  identifier, so a dashboard built on it selects nothing;
- ADR-029 §4 names "caller and tenant attributes" from the bound identity
  context, and session is in the same family. A field that is always `None` is
  not a privacy-preserving omission, it is a gap.

## Where

`CallerIdentity.session_id` (`domain/value_objects/identity.py`) is populated by
whoever builds the context. The API-key path does not set it. Worth checking
whether the OIDC and header-identity paths do, and whether the SDK exposes the
session id to the ASGI layer where `bind_caller_identity` runs at all -- if it
does not, this is a question about which identifier is honest to use, not a
missing assignment.

## Out of scope

Tracing. #1278 is complete without this: the boundary omits what it is not
given, which is the behaviour the contract asks for. This issue is about the
identity context being given less than the request carries.

Issue #1293, opened 2026-09-10
test(core): verify trace and audit export through a live gateway


mapyr:
## Goal

Prove black-box that a real gateway exports traces and audit records that arrive at an OTLP receiver, including across a graceful shutdown.

## Symptom and evidence

Report sections 8-9 (live verification): in-process and in-memory tests cannot prove export, startup configuration, shutdown delivery or front-door wiring. At bd6a62c9:

- `docs/internal/LIVE_VERIFICATION.md:63` still marks trace propagation through a collector as proven only with a mocked context.
- `tests/live/README.md:19` puts the observability tier (T3) out of scope. The live tier is opt-in, not a per-PR gate (`:7`); `.github/workflows/live-verify.yml` runs it nightly and on dispatch (`:7-14`) with all extras and `MCP_HANGAR_LIVE_VERIFY=1` (`:36-41`).
- `live_http_hangar` starts `serve --http` as a subprocess (`tests/live/conftest.py:95-96`), so OTLP env can be injected.
- Audit records are dropped today (#1289). `service.instance.id` given in `OTEL_RESOURCE_ATTRIBUTES` reaches the resource unchanged (reproduced with SDK 1.44), so it can carry a per-run ID.

Not verified: that Docker is usable for a Collector container in the live job on `ubuntu-latest`.

Audit baseline: `ff3167344af27e77149681ca16fb66261ce34711` (2026-09-10). Re-verified against main `bd6a62c9d818b9dffa4f1f5c1c9af486d427b2fa` on 2026-09-10: since the baseline, `src/` changed only in `metrics.py`, `observability/conventions.py`, `observability/__init__.py` and docstrings in `domain/events/analysis.py` (#1261, #1263). Repository evidence with local reproductions as described above; not verification of a live installation.

## Proposed change

Add a T3 export smoke test. Start a real OTLP receiver (the #1291 Collector config with a file exporter, or an in-process OTLP/HTTP receiver), start the gateway pointed at it with a unique `service.instance.id`, run one warm call and one denied call, then send SIGTERM. Assert at the receiver with bounded polling. Update the tier table and the LIVE_VERIFICATION row to what is proven.

Split off:

[...]

Issue #1266, opened 2026-09-10
feat(core): make governance decisions explainable from traces


mapyr:
## Background

Delivery story under https://github.com/mcp-hangar/mcp-hangar/issues/1264, based on the 2026-09-10 core tracing audit.

## Why

An operator can identify the gate, authoritative reason, policy revision and selected backend for a call.

## What changes

Implement the individually scoped child tasks. This story is an aggregation, not a single implementation PR.

## Child issues

**P1**

- [ ] #1275 — feat(core): define the additive tracing and decision contract
- [ ] #1276 — chore(security): review the telemetry data handling contract
- [ ] #1278 — feat(core): enrich governance spans on the active command path
- [ ] #1285 — feat(core): expose batch gate outcomes and decision reasons

**P2**

- [ ] #1286 — feat(core): explain logical target and backend selection in traces
- [ ] #1295 — feat(core): expose authoritative L7 enforcement verdicts in traces
- [ ] #1297 — feat(core): trace command and query dispatch across management entry points

## Suggested order

Follow the explicit child dependency graph. Isolated correctness fixes do not wait for an architecture redesign.

## Acceptance criteria

- [ ] The agreed additive contract is emitted on real paths, with maintainer-reviewed data handling and no enforcement behavior changes.
- [ ] Child implementation and validation evidence reviewed individually.

## Out of scope

Unrelated runtime behavior changes, production deployment and cross-repository monitoring ownership changes.

## References

https://github.com/mcp-hangar/mcp-hangar/issues/1264

mapyr:
## Story analysis — 2026-09-10

This story contains 7 individually reviewed tasks. It is not a single implementation PR.

| Task | Priority | Ownership | Prerequisites |
|---|---|---|---|
| #1275 | P1 | maintainer | none |
| #1276 | P1 | maintainer | none |
| #1278 | P1 | agent-suitable | #1275, #1276, #1270 |
| #1285 | P1 | maintainer | #1275, #1276, #1278 |
| #1286 | P2 | maintainer | #1275, #1276, #1278 |
| #1295 | P2 | maintainer | #1275, #1276, #1278 |
| #1297 | P2 | agent-suitable | #1275, #1276, #1270 |

The dependency graph separates immediate defect repairs from contract-dependent additions. Keep gate order, existing attribute compatibility and private security handling intact. Completion requires the story acceptance criteria plus evidence from the individual tasks, not a count of newly added spans.

[...]

Issue #1264, opened 2026-09-10
epic: make core tracing explain execution and governance decisions


mapyr:
## Background

The 2026-09-10 audit of core at ff316734 found partial OpenTelemetry instrumentation, with locally reproduced parentage and upstream-status defects and missing decision context. Selected real-SDK tests passed (97), demonstrating a contract-coverage gap rather than absence of tracing. The live installation itself was not inspected.

Re-verification against main `bd6a62c9` on 2026-09-10 corrected one premise. The mcp 2.0 SDK already opens a SERVER span parented from `params._meta` (`OpenTelemetryMiddleware`, on by default; Hangar does not disable it), so over streamable HTTP one request is one trace. The parentage defect that remains is `batch.call` re-parenting past the local `hangar_call`/`batch.execute` spans (#1270). The re-verification also found that core CI installs no OpenTelemetry SDK, so every real-SDK test in this epic skips until #1269 lands.

## Why

Operators must be able to explain target selection, refusal, waiting, retries and asynchronous effects without reconstructing implementation code during incidents.

## What changes

Four delivery stories: request correctness; decision visibility; lifecycle/causality; OTLP delivery and operational validation. Child issues define atomic goals, evidence, acceptance criteria, implementation limits and dependencies.

## Versioning impact

Preserve existing trace names/attributes additively. Do not introduce a major release. New architecture decisions require maintainer-authored ADRs; small defect repairs can proceed independently.

## Child issues

- [x] #1265 — feat(core): make request traces structurally and semantically reliable
- [ ] #1266 — feat(core): make governance decisions explainable from traces
- [ ] #1267 — feat(core): explain lifecycle latency and asynchronous causality
- [ ] #1268 — feat(core): verify telemetry delivery and operational usability

## Suggested order

Priorities: P1 for 16 tasks (#1269-#1278, #1283, #1284, #1285, #1289, #1291, #1293), P2 for 18, and P3 for 3 (#1290, #1302, #1305).

**Maintainer track** (starts now, in parallel with the implementation track):

1. #1275 (contract) gates 13 tasks. #1276 (data-handling review) gates 7; the private advisory for the related finding is drafted. The contract is the biggest lever: until it is accepted, most of the decision and lifecycle work cannot start.
2. Later:
   - #1302 (renderer decision), before #1290;
   - the human-required implementations #1285, #1286, #1288 and #1295, once #1275 and #1276 are accepted;
   - #1303 (overhead budget sign-off), last.

[...]

PR #1540, opened 2026-09-21
feat(core): put the caller's identity on the governance span

### LINKED ISSUE #1278: feat(core): enrich governance spans on the active command path
## Goal

Emit approved governance context on real tool invocations without relying on an unconstructed service decorator.

## Symptom and evidence

F06: `TracedMcpServerService` (`application/services/traced_mcp_server_service.py:114-162`) sets governance attributes and risk scores, but nothing in `src/` constructs it; only tests do. Both real entries run `BatchExecutor.execute`: `hangar_call` (`server/tools/batch/__init__.py:570-576`) and flat tools (`fastmcp_server/flat_tool_projection.py:928-944`). Per call: `batch.call.{tool}` (`server/tools/batch/executor.py:1034-1053`), `command.send.InvokeToolCommand` (`:1570-1585`), `handler.InvokeToolCommand` (`infrastructure/command_bus.py:129-132`); `InvokeToolHandler` then calls the aggregate directly (`application/commands/handlers.py:142-148`). An `execute_tool {tool}` CLIENT span already exists per upstream RPC (`observability/tracing.py:411-435`); the mcp 2.0 SDK already opens a SERVER span `tools/call {tool}` per message (`mcp/server/lowlevel/server.py:439`), not disabled by Hangar.

Identity: the executor reads `get_identity_context()` (`executor.py:1092`). Over streamable-HTTP the ASGI-bound contextvar does not reach tool handlers; it is re-bound by `hangar_call` (`batch/__init__.py:548-557`) and on the SDK v2 flat path (`flat_tool_projection.py:1006-1016`). The flat path passes no `request_ctx` to the executor. Risk signals are recorded only by `RiskScoringEventHandler` (`server/bootstrap/event_handlers.py:200-201`).

Checked: reference search and static reading at bd6a62c9. Not verified: identity on the SDK v1 flat path; exported attributes from a live installation; the body of closed #106.

Audit baseline: `ff3167344af27e77149681ca16fb66261ce34711` (2026-09-10). Re-verified against main `bd6a62c9d818b9dffa4f1f5c1c9af486d427b2fa` on 2026-09-10: since the baseline, `src/` changed only in `metrics.py`, `observability/conventions.py`, `observability/__init__.py` and docstrings in `domain/events/analysis.py` (#1261, #1263). Repository evidence with local reproductions as described above; not verification of a live installation.

## Proposed change

Enrich the boundary #1275 names (expected: the existing `batch.call.{tool}` span, shared by batch and flat tools) with the agreed attributes, read once from authoritative sources. No second `execute_tool` span; do not revive the decorator as a parallel pipeline.

## Files in scope

- `src/mcp_hangar/server/tools/batch/executor.py`
- `src/mcp_hangar/application/commands/handlers.py`
- `src/mcp_hangar/infrastructure/command_bus.py` (if #1275 picks the handler span)
- `src/mcp_hangar/observability/conventions.py`
- `src/mcp_hangar/application/services/traced_mcp_server_service.py` (removal only, per #1275)
- `tests/live/`
- `tests/integration/`

## Files out of scope

Unrelated product changes, release machinery, version bumps, existing trace-attribute renames, production deployment. Protected security/auth middleware and the identity bridges themselves. No new approval surfaces, changes to gate precedence, or duplicate monitoring stack.

## Acceptance criteria

- [ ] `hangar_call` and flat-tool invocations emit the #1275 attribute set on one span per logical call.
- [ ] Caller, tenant and session attributes come only from the bound identity context; unknown values omitted, never read from baggage.
- [ ] No second `execute_tool` span; existing span names and attributes unchanged.
- [ ] Risk and cost not recomputed; no duplicate risk-signal or cost side effects.
- [ ] Tracing disabled: no behavior change, no exception.
- [ ] `tests/live` over streamable-HTTP with an authenticated tenant shows identity-derived attributes on a `hangar_call` and a flat-tool call; a mock-ctx unit test does not satisfy this.
- [ ] Decorator deleted, or its retention recorded, per #1275.

## Max LOC declared

At most 350 non-test implementation LOC in one PR; tests are budgeted separately per repository guidance. Split before implementation if this estimate is exceeded.

## Single Conventional Commit scope

`core`. Agent-suitable only within the declared file scope and LOC cap, and only after #1275 lists exact attributes and sources; one goal and one Conventional Commit scope.

## Dependencies

Blocking:
- https://github.com/mcp-hangar/mcp-hangar/issues/1275 — attribute set, enrichment boundary, decorator decision.
- https://github.com/mcp-hangar/mcp-hangar/issues/1276 — which identifiers may be exported.

Ordering (not blocking):
- https://github.com/mcp-hangar/mcp-hangar/issues/1270 — keeps `batch.call` spans under their local parents; enrich after.
- https://github.com/mcp-hangar/mcp-hangar/issues/1269 — SDK tests must run in CI.
- https://github.com/mcp-hangar/mcp-hangar/issues/1293 — its export smoke gives the live tier a span sink; the `tests/live` criterion here needs it to assert spans.

## Parent story

https://github.com/mcp-hangar/mcp-hangar/issues/1266

## Severity

P1 — Next-release tracing correctness or essential decision visibility.

## Implementation constraints

Shares `infrastructure/command_bus.py` with #1297, which lands after this task.

Risk and cost must not be computed twice or inferred from labels; identity is not trusted merely because it appears in baggage.

Add one changelog.d fragment for the implementation PR; do not edit CHANGELOG.md. No major-version commit footer.

## References

- https://github.com/mcp-hangar/mcp-hangar/blob/bd6a62c9d818b9dffa4f1f5c1c9af486d427b2fa/src/mcp_hangar/server/tools/batch/__init__.py#L548-L576
- https://github.com/mcp-hangar/mcp-hangar/blob/bd6a62c9d818b9dffa4f1f5c1c9af486d427b2fa/src/mcp_hangar/fastmcp_server/flat_tool_projection.py#L928-L1016
- https://github.com/mcp-hangar/mcp-hangar/blob/bd6a62c9d818b9dffa4f1f5c1c9af486d427b2fa/src/mcp_hangar/server/tools/batch/executor.py#L1034-L1092

mapyr:
## Why

`set_governance_attributes` had exactly one caller, `TracedMcpServerService`,
and nothing in `src/` ever constructed it. The governance attributes ADR-007
describes were therefore emitted by tests and by nothing else, while every real
invocation — `hangar_call` and every front-door flat call alike — produced a
`batch.call.<tool>` span that could not say who made it.

ADR-029 §4 names that span as the one enrichment boundary and retires the
decorator rather than wiring it in, because wiring it would have rebuilt a
second pipeline beside the one that runs.

Closes #1278. Parent story #1266, epic #1264.

## What

- `batch.call.<tool>` sets `mcp.caller.type`, `mcp.caller.id`,
  `mcp.caller.tenant_id`, `mcp.user.id`, `mcp.agent.id`, `mcp.session.id` and
  `mcp.correlation_id`, read once from the bound identity context.
- Unknown values are **omitted**, not exported empty.
- **Baggage is not a source.** Anything that forwards a request can write it, so
  a tenant label taken from there would be a claim this gateway never
  authenticated. A unit test holds that line.
- `TracedMcpServerService` and its `TracedProviderService` alias are deleted,
  with an `upgrade.d/` note.
- `set_governance_attributes` is **kept**, and its docstring now records why the
  boundary does not call it (below).
- The two existing literals on that span become `conventions.py` constants, per
  ADR-029 §9.

## Why the boundary does not reuse `set_governance_attributes`

That helper also asserts `gen_ai.operation.name=execute_tool` and
`mcp.method.name=tools/call`, which name the **upstream** call. ADR-029 §3 keeps
those on the one CLIENT span `execute_tool <tool>`; repeating them on the
governance span would invite a GenAI-aware backend to count one invocation
twice. The function stays because it is public surface an ADR-007 adapter may
call with its own process-local invocation data.

## What the live run found

An API-key caller over streamable-HTTP has **no session id** in its
`IdentityContext`, so `mcp.session.id` is omitted:

```
batch.call.add  mcp.caller.id=svc:tenant-trace  mcp.caller.tenant_id=tenant-trace
                mcp.caller.type=service         mcp.user.id=svc:tenant-trace
```

That is the contract working — an unknown value is omitted — but it is absent
because `CallerIdentity.session_id` is `None`, not because the request had no
session. The live test asserts the **absence**, with a comment saying to move
the key into `_REQUIRED` and delete the assertion when it changes, so the day it
does is a red test rather than a silent one. Filed as #1539; it is identity
plumbing, not tracing, and #1278 is complete without it.

## How tested

- `tests/live/test_t3_governance_identity.py`, **2 passed** — two real
  `mcp-hangar serve --http` processes (one `egress` for `hangar_call`, one
  `front_door` for a flat tool, because neither surface serves the other's entry
  point), a real `X-API-Key` carrying a tenant, attributes read back out of OTLP
  protobuf at the receiver from #1293. The acceptance criteria require this tier
  explicitly and say a mock-ctx unit test does not satisfy it — for a mechanical
  reason: identity reaches a handler through an ASGI contextvar the SDK's
  per-message task does not inherit, and each surface re-binds it.
- `tests/unit/observability/test_governance_span_identity.py`, 6 passed — the
  mapping, omission of unknowns, the baggage refusal, and the ADR-026 stdio
  fallback.
- `uv run pytest tests/unit` — **9363 passed**, 3 skipped.
- `uv run pytest tests/integration` — **756 passed**, 7 skipped.
- `uv run mypy src/` — clean, 442 files. `ruff check` and `ruff format --check`
  clean.

## Acceptance criteria

- [x] `hangar_call` and flat-tool invocations emit the attribute set on one span per logical call
- [x] Caller, tenant and session attributes come only from the bound identity context; unknown omitted, never from baggage
- [x] No second `execute_tool` span; existing span names and attributes unchanged
- [x] Risk and cost not recomputed; no duplicate risk-signal or cost side effects (neither is read here)
- [x] Tracing disabled: no behaviour change, no exception — the helper reads a contextvar and no OTel state
- [x] `tests/live` over streamable-HTTP with an authenticated tenant, on both surfaces
- [x] Decorator deleted, per ADR-029

## Risk and rollback

Additive on the span. The one removal is public API nothing in the product used,
covered by the upgrade note. Revert is the single commit.

## Agent metadata

- [x] LOC declared upfront: `~40 non-test implementation LOC (cap was 350); the diff is +49/-883 because four test files covered the deleted class`


mapyr:
**CI went red once on something this PR does not touch, and is green on rerun.**

`test_held_calls_leave_threads_and_descriptors_at_baseline[cancelled_batch]`
failed on `test (3.11)` with `assert 'TimeoutError' == 'approval_timeout'`. That
scenario sets a 0.2 s batch deadline and a 1 s approval hold and asserts which
of the two names the failure; on a loaded runner the global timeout surfaced
instead. The two deadlines are 800 ms apart and this PR adds one contextvar read
and a seven-entry dict per call, so it cannot move that race. `test (3.14)` was
`cancelled` by fail-fast, not failed, so it was not a second data point.

Evidence:

- the same commit, rerun: `CI - Core` green, 28/28 checks pass;
- locally with this change: that test 3/3, `tests/unit` 9363 passed,
  `tests/integration` 756 passed.

Filed as #1541 rather than left as "flaky CI": the flake is the visible end of
an unanswered question -- when a held call's batch deadline and its approval
deadline both pass, which one is the caller told about? The two answers send an
operator to different places.

